PR #2, opened 2014-11-14
[REF] auto-fix conventions

moylop260:


clonedagain:
:+1: ok why not

pedrobaeza:
It doesn't apply. Branch is already green and the rule hasn't been correctly applied.

PR #11, opened 2016-03-08
Add code3166 3-char codes to res_country

elacunza:
This is a initial PR to add 3-char codes to res_country, needed for electronic invoices in Spain.

I suspect existing 2-char codes in res_country are also ISO 3166-1, so maybe we should fix module and new field names. Also maybe iso3166_1 would be a better name than code3166?

@pedrobaeza feel free to suggest changes or to make them directly, I won't be able to look into this until tomorrow morning CET.

Thanks

### DIFF
--- a/code3166/README.rst
+++ b/code3166/README.rst
@@ -1,0 +1,47 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========
+ISO 3166-1
+==========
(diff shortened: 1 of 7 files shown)


pedrobaeza:
The name should be _base_country_iso_3166_. Anyway, don't worry about that, because I will take your commit as start point and propose other PR with the rest.

Thanks for the work.

elacunza:
Thanks, feel free to close this PR; otherwise comment additional changes so that I can work on them tomorrow.

pedrobaeza:
Superseeded by https://github.com/OCA/community-data-files/pull/12

PR #12, opened 2016-03-08
[8.0] [ADD] base_iso3166

pedrobaeza:
# 
# ISO 3166 adaptation to Odoo

This module adds some data specified by ISO 3166, specifically:
- ISO 3166-1 alpha-3 (3-char) codes in countries
- ISO 3166-1 numeric codes in countries

Check https://en.wikipedia.org/wiki/ISO_3166 for more info.
# Installation

This module requires the Python library pycountry installed in the system.
# Known issues / Roadmap
- Include 3166-2 country subdivision codes
- Include 3166-3 formerly used names of countries codes


pedrobaeza:
cc @elacunza, @antespi, @Yajo, @omar7r, @rafaelbn 
Needed for spanish Factura-e

bealdav:
@pedrobaeza I had to negociate with these data too. From my side I used pycountry  https://pypi.python.org/pypi/pycountry

Here is an example https://github.com/OCA/carrier-delivery/blob/7.0/delivery_carrier_label_gls%2Fstock.py#L106

and another: https://github.com/akretion/logistics-center/blob/7.0/bleckmann_logistics%2Foutput.py#L152

Advantage to use an existing library: 
- this is the common way for python world and debian
- copy/paste legal info can lead to have difficult up to date information in entire python ecosystem: think tryton, erpnext, odoo only for erp world
- give you, some time with your family: someone else maintain that for you
- save bytes

But in your module, its' good idea to display this info in country view: then you can make an computed field to display information in the view from pycountries.

What do you think ?

pedrobaeza:
It seems to me a very good idea indeed. Let me work on that then.

pedrobaeza:
@bealdav, done!

bealdav:
Yeah, you saved 700 lines of xml data in 20 minutes. You're a champion. ;-)

bealdav:
I don't know if we need it, but this module could be more generic for few lines more

https://en.wikipedia.org/wiki/ISO_3166-1_numeric

pedrobaeza:
Thanks for the tip. I was making things Odoo XML style...

pedrobaeza:
Well, for the numeric code, we then need to rename the module and the scope will change. I prefer to do that in another module.

base_country_iso3166_1_alpha3/models/res_country.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @api.multi
+    @api.depends('code')
+    def _compute_code_alpha3(self):
+        for country in self:
+            try:
+                country.code_alpha3 = pycountry.countries.get(
+                    alpha2=country.code).alpha3
+            except KeyError:
+                country.code_alpha3 = ""
```
bealdav: proposal: 
log message "please consider to update pycountry lib"

pedrobaeza: I prefer not, because Odoo includes obsolete countries like "Yugoslavia" that aren't included in the library, and you won't get it although you update the library. There's also the possibility of adding or modifying manually a country, putting an incorrect code.

bealdav: you're right, ok for me

bealdav:
Take the time to think about rename module and extend to numeric value.

base_country_iso3166 or country_iso3166 ?

Ratio  python / README will be better ;-)

Really nice for your work, as usual.

:+1:

legalsylvain:
Thanks @pedrobaeza 
:+1:
(code review + functional test).
Possible improvement : improve search function. (for the time being, MDG doesn't give any results, and could return Madagascar).
don't you think ?

ivantodorovich:
:+1: 

I agree with @bealdav, it's better to have just one `base_country_iso3166` that supports alpha3 and numeric than two separate modules with a repeated code base. In just one pycountry api call you can populate both alpha3 and numeric codes.

Also agree with @legalsylvain regarding the search function, seems straightforward. :+1:

base_country_iso3166_1_alpha3/models/res_country.py
ivantodorovich: You could improve it to query the "Historic Countries Database" when no results are found on the first try.

> https://pypi.python.org/pypi/pycountry 
> The historic_countries database contains former countries that have been removed from the standard and are now included in ISO 3166-3, in addition to the existing ones:

``` py
pycountry.historic_countries.get(alpha2=country.code)
```

And maybe do show a log output warning, as @bealdav suggested, if no results are found in either two calls. At this point, it's either a new country or the user entered a bad country code. Either way, a log warning seems usefull.

alexis-via:
@pedrobaeza out-of-topic: I read some specs of Factura-e (as you know, I'm very hot on electronic invoicing these days !) ; I read those: http://www.facturae.gob.es/formato/Documents/EspanolFacturae3_0.pdf

It seems the XML syntax is specific to spain. Are you supposed to send the XML embedded in the PDF of the invoice, or are you suppose to send only the XML file ? The spec I read didn't speak about that... Anyway, it should be pretty easy to implement a module account_invoice_import_facturae just like account_invoice_import_zugferd :)

We can continue the discussion elsewhere if you don't what too much off-topic discussion on this PR.

rafaelbn:
Thanks :+1:

.travis.yml
```diff
@@ -6,6 +6,7 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
+      - python-pycountry
```
yajo: Shouldn't this go into requirements.txt?

base_country_iso3166_1_alpha3/__openerp__.py
```diff
@@ -0,0 +1,31 @@
...
+
+{
+    "name": "ISO 3166-1 alpha-3 codes for countries",
+    "version": "8.0.1.0.0",
+    "author": "ASR-OSS, "
+              "Comunitea, "
+              "Binovo IT Human Project, "
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Odoo Community Association (OCA)",
+    "category": "",
```
yajo: Add something here. `Tools` for example.

base_country_iso3166_1_alpha3/models/res_country.py
```diff
@@ -0,0 +1,30 @@
...
+
+    code_alpha3 = fields.Char(
```
yajo: Please follow [guidelines](https://www.odoo.com/documentation/8.0/reference/guidelines.html#symbols) and put fields above methods.

base_country_iso3166_1_alpha3/models/res_country.py
```diff
@@ -0,0 +1,30 @@
...
+        string='Country Code (3-letter)', size=3, store=True,
+        help='ISO 3166-1 alpha-3 (three-letter) code for the country',
+        compute="_compute_code_alpha3")
```
Yajo: Add `store=True`.

base_country_iso3166_1_alpha3/models/res_country.py
yajo: I agree with @ivantodorovich, and I add that empty value must be `False`, not `""`.

base_country_iso3166_1_alpha3/models/res_country.py
pedrobaeza: Thanks, I always change the convention because in v7 compute methods go above

.travis.yml
pedrobaeza: It can be in both places, but I'll do in requirements.txt to specify the minimum version where historic countries has been introduced

base_country_iso3166_1_alpha3/models/res_country.py
yajo: I know, but this is 8.

.travis.yml
bealdav: Should we maintain 2 list for ever, or just one is sufficient ?

yajo: 1 better.

pedrobaeza:
Thanks to all the reviewers. I didn't expect to have this interest!

I have expanded the module to host all the ISO 3166 specification, although the country subdivision is not implemented (mentioned on the known issues), but alpha-3 and numeric codes are both included.

@alexis-via, about Factura-e, is an XML file signed with a clumsy certificate, but XML at the end, so yes, an account_invoice_import extension can be easily made and it would be certainly very welcome. Anyhow, as the Factura-e is only mandatory for invoices made to the government, don't expect to see this accepted in regular B2B invoices.

hbrunn:
:+1:

PR #13, opened 2016-03-10
[ADD] base_bic_from_iban

hbrunn:
As this is mostly data, I think this belongs here. I'd also rather have data for all countries here as opposed to put that into the according localization repos because then you would have to collect all of them in order to support i.e. all European countries.


pedrobaeza:
Well, I don't agree of having all the information centralized instead of having on each localization, because this means that the migration/update of this module to future versions will need a lot of hands to be completed and the modularity will be lost.

hbrunn:
Then supporting Europe would be having to add 50+ localizations, do you really think that's feasible?

About migrations: For countries where it's only data, no special knowledge would be required. I can agree on having countries that implement the lookup with code in `lookup_bic_XX` in the appropriate localization, can we meet there maybe?

pedrobaeza:
For example, in Spain the module for bank BICs retrieves the file from Internet, parses it and create records. That's why I say it's a lot to maintain in only one module for all localizations:

https://github.com/OCA/l10n-spain/tree/8.0/l10n_es_partner

hbrunn:
That would be a good case for having this in l10n-es. What do you think about the needing-a-lot-of-modules-for-pan-European-installations-issue?

pedrobaeza:
If we would be able to provide an unique data-source, like in module _base_location_geonames_import_, to have one module would be the ideal, but having multiple sources... I prefer not.

The concept of Odoo is the modularity, and we already installed modules that add one field in one model, why not one for each handled country?

Do you know if the European bank has any source for consulting at least all European banks?

hbrunn:
yes, that exists: http://swiftref.swift.com/distribution-channels

But from my understanding, you need to be involved in organized crime (aka financial institutes) to be allowed to sign up there in the first place and it doesn't look exactly cheap either

pedrobaeza:
Ouch, what happens with EU open data?! :disappointed:

hbrunn:
that only applies to government agencies, not to organized crime unfortunately

pedrobaeza:
What do we do then? I'm going to tell people in IRC to give their opinion.

ivantodorovich:
I don't have a well-formed opinion, but I wanted to include a non-european point of view into the discussion.

There are already some modules that only applies to european countries in general repositories. For instance:
- https://github.com/OCA/bank-payment: account_banking_sepa_%
- https://github.com/OCA/partner-contact: base_location_nuts

It seems localization rules applies well to country-specific modules, but there are some modules that are used in more than one european country and they are among the "global" modules.

Don't take me wrong, I'm not complaining about that. Just wanted to point out that It seems the rules aren't clear enough in these cases.

dreispt:
I feel that this is rather l10n_nl specific.
But you could a "bundle module" that installs all l10n specific modules providing BIC codes, wither here or in a bank related repo.

hbrunn:
My intention with this module is to centralize the lookup process, so that we don't get competing lookups. Where the data lives then doesn't matter that much in the first instance. But can we agree it's a good idea to have a general module here that is extended by other ones, be it in some localization or not?

pedrobaeza:
Yeah, for me it's a good idea to have this module. The project for this one is difficult to set...

gurneyalex:
:+1:

pedrobaeza:
I have several concerns about this module:
- NL data should be in a localization module that depends on this one.
- For Spain, we only need to check 4 digits of the IBAN number, and match with the bank code, and thus, the bank is matched and it contains the BIC. With this module, we will need to duplicate the data, one for the bank itself, and another for matching the BIC.

hbrunn:
@pedrobaeza for your first point, I can move them away if everyone else agrees this should happen, while I still think it would be more convenient for users to have all of this in one module.

For your second point: Have a look at https://github.com/OCA/community-data-files/pull/13/files#diff-d071221eed5f4021a008161548e7ab27R31 - it's that whole point of this module that every country can have its own computation function.

So if I understand you correctly, the only thing Spain would need would be

``` python
class ResBankIbanBicMapping(models.Model):
    _inherit = 'res.bank.iban.bic.mapping'

    def lookup_bic_ES(self, iban):
        return iban[4:8]
```

or the like, isn't it? The whole idea with this module is that I want to propose streamlining the lookup and centralize it, so that not every localization has to cook up something for its own.

bosd:
> yes, that exists: http://swiftref.swift.com/distribution-channels
> 
> But from my understanding, you need to be involved in organized crime (aka financial institutes) to be allowed to sign up there in the first place and it doesn't look exactly cheap either

This might be a very interesting data source for BIC data.
https://www.iso9362.org/isobic/overview.html

In PDF format (be patient, it is loading).
https://www.iso9362.org/bicservice/public/v1/bicdata/_pdf

PR #14, opened 2016-03-21
8.0 Add UNECE modules

alexis-via:
These modules are required for electronic invoicing (CII, ZUGFeRD, UBL, etc...)

### DIFF
--- a/account_payment_unece/__openerp__.py
+++ b/account_payment_unece/__openerp__.py
@@ -1,29 +1,11 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Account Payment UNECE module for Odoo
-#    Copyright (C) 2016 Akretion (http://www.akretion.com)
-#    @author Alexis de Lattre <alexis.delattre@akretion.com>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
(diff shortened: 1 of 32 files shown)


alexis-via:
Runbot doesn't take into account the file  oca_dependencies.txt ; that's why runbot is red. But travis takes it into account already, that's why it is green.

pedrobaeza:
Dependency added on runbot. Are you sure this project is the best suited for this module? I think _account-invoicing_ is a better host.

alexis-via:
@pedrobaeza I don't know which project is the most suited. I thought it was this one, because there are a LOT of unece codifications and what I propose is only a small part of it, the part which i need for CII and UBL invoices... but we may need other UNECE codifications in the future of other needs.

account_payment_unece/__openerp__.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Use short headers like https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

account_payment_unece/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': 'Account Payment UNECE',
+    'version': '8.0.0.1.0',
```
pedrobaeza: 8.0.1.0.0

pedrobaeza:
@alexis-via, you're right that the project is very transversal, so it can fit here. As I was thinking about your account_invoice_import modules, I said the other, but this is bigger and different than that.

gurneyalex:
:+1:

bguillot:
:+1:

alexis-via:
Can someone with the super-powers merge this please ?

alexis-via:
@pedrobaeza I know you have the commit rights on all OCA projects, could you merge this please ?

pedrobaeza:
@alexis-via, sorry for not attending on time, but @sebastienbeau has made it instead of me. Thanks!

PR #15, opened 2016-09-12
8.0 Update account_payment_unece

alexis-via:
Add payment means 49 (direct debit) with some explainations. The equivalent data has been removed from the base_zugferd module.


Doscaal:
:+1:

pedrobaeza (approved):

PR #21, opened 2017-01-06
Update README.rst

bealdav:
BREAKING API in lib:

This module requires the Python library pycountry >= 0.19 (and < 16.10.23rc1) installed

ping @pedrobaeza

### DIFF
--- a/base_iso3166/README.rst
+++ b/base_iso3166/README.rst
@@ -17,7 +17,7 @@ Check https://en.wikipedia.org/wiki/ISO_3166 for more info.
 Installation
 ============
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Better to adapt the module than to change the README

bealdav:
You're right but no time for that for now around me

but it's better to alert than nothing.

sorry

bealdav:
Any news, there.

pedrobaeza:
I still think the adaptation is very quick (just a `getattr` call with an if). Can you try?

bealdav:
So many PR awaiting, sorry, I couldn't. So you can close if you want

PR #24, opened 2017-06-01
[10.0] account_payment_unece: add UNECE codes on data from 'account' module

alexis-via:
A small PR to ease tests on the OCA/edi project.

### DIFF
--- a/account_payment_unece/__manifest__.py
+++ b/account_payment_unece/__manifest__.py
@@ -14,6 +14,7 @@
     'depends': ['account_payment_mode', 'base_unece'],
     'data': [
         'data/unece.xml',
+        'data/account_payment_method.xml',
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

pedrobaeza:
I prefer to put that method on a base test here that will be inherited on the each module. See an example here.

Base class: https://github.com/OCA/l10n-spain/blob/9.0/l10n_es_aeat/tests/test_l10n_es_aeat_mod_base.py#L13
Inherited class that calls methods from base: https://github.com/OCA/l10n-spain/blob/9.0/l10n_es_aeat_mod303/tests/test_l10n_es_aeat_mod303.py#L12

pedrobaeza:
Thanks!

PR #31, opened 2017-10-31
fix(base_iso3166): add codes of countries according to the code iso 3166

agb80:
modified res_country.py with the adjustments made for version 10 since it did not recognize the
two-letter code when inserting it for what it returned false. with this adaptation already shows the
codes to three characters according to the standard Iso 3166

### DIFF
--- a/base_iso3166/README.rst
+++ b/base_iso3166/README.rst
@@ -1,5 +1,5 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :target: https://www.gnu.org/licenses/agpl-3.0-standalone.html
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Please cherry-pick the commit and adapt anything remaining instead of hijacking the change with your own commit.

agb80:
@pedrobaeza done

aitorbouzas:
@agb80 would you mind increasing the code coverage so this module can be merged?

agb80:
I could add a test file that check code for one or two countries, is that ok?

yajo (approved):

pedrobaeza:
> I could add a test file that check code for one or two countries, is that ok?

Yes, that should be enough

pedrobaeza:
Merging as old one with at least the cherry-picking comment for having this in v8.

PR #32, opened 2017-11-15
[11.0][NEW] base_bank_from_iban: New module

carlosdauden:
This module adds code to bank and fill bank from IBAN in Bank Account

@Tecnativa


base_bank_from_iban/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+To use this module, you need to:
+
+#. Go to Partner
+#. Click Bank Account(s) in Sales & Purchases tab
+#. Create/modify IBAN bank account
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/110/10.0
```
yajo: 11.0

base_bank_from_iban/models/account_journal.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 Tecnativa - Carlos Dauden <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3).
+
+from odoo import api, models
+
+
+class AccountJournal(models.Model):
+    _inherit = 'account.journal'
+
+    @api.multi
```
yajo: Implicit since odoo 10.0

base_bank_from_iban/models/res_bank.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2017 Tecnativa - Carlos Dauden <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3).
+
+from odoo import fields, models
+
+
+class ResBank(models.Model):
+    _inherit = 'res.bank'
+
+    code = fields.Char('Code')
```
yajo: The field string equals the name, so it's implicit. You can remove it.

base_bank_from_iban/models/res_partner_bank.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo import api, models
+from odoo.addons.base_iban.models.res_partner_bank import \
+    normalize_iban, pretty_iban, _map_iban_template
+
+
+class ResPartnerBank(models.Model):
+    _inherit = 'res.partner.bank'
+
+    @api.multi
+    @api.onchange('acc_number')
```
yajo: According to the code, this should depend on `acc_type` too

base_bank_from_iban/models/res_partner_bank.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import api, models
+from odoo.addons.base_iban.models.res_partner_bank import \
+    normalize_iban, pretty_iban, _map_iban_template
+
+
+class ResPartnerBank(models.Model):
+    _inherit = 'res.partner.bank'
+
+    @api.multi
```
yajo: Implicit

yajo (changes requested):
Looks awesome, pal. Just some minor improvements.

base_bank_from_iban/models/res_partner_bank.py
carlosdauden: It shouldn't be possible to change acc_type without changing acc_number

pedrobaeza: The problem is that we don't know if `acc_type` is computed before or after calling this onchange, so it's better to include that field as dependency here for not depending on the ORM implementation.

carlosdauden:
@Yajo @pedrobaeza changes done

yajo (approved):

hbrunn:
did you check https://github.com/OCA/community-data-files/pull/13 ? I think the mechanism there is a bit more general, and what's done here can be very simply made fit

pedrobaeza:
@hbrunn I think this mechanism is more standard (it's even an ISO) and NL can fit also here.

pedrobaeza (approved):

pedrobaeza:
I'm going to merge this one as v11 includes support for this directly and it's a dependency from other modules. @hbrunn you can extend this one for your programatic approach. In Spain, that approach can't be and we need to hardcode on each bank the correspondent code.

PR #39, opened 2018-06-21
[11.0][MIG] l10n_eu_nace: Migration to 11.0

oihane:


pedrobaeza:
We talked about changing partner labels to specific fields. Have you done anything in this way?

legalsylvain:
@oihane : seems that the removal of data is a problem. ref https://github.com/OCA/community-data-files/pull/53#issuecomment-450632863. 
what do you think ? CC : @sbejaoui

regards.

sebalix:
Better to focus on https://github.com/OCA/community-data-files/pull/60 I think which doesn't mix NACE with partner tags.

remi-filament:
@sebalix this PR also creates a specfic table for NACE codes, from my point of view #60 and #39 are redundant...
@pedrobaeza Which one to go for then (this #39 seems to include some of the comments from @Yajo on #60) ?

I installed that one on V12 (with a couple of changes in code, like creating parent_path field in res_partner.py), and indeed l10n_fr_naf_ape will need refactoring since external ids for NACE are not available anymore...

PR #41, opened 2018-09-29
[12.0][MIG] base_bank_from_iban: Migration to 12.0

pedrobaeza:
Standard migration

cc @Tecnativa


pedrobaeza:
Fast-tracking as OCB tests are green and this is a requirement for other modules.

Issue #44, opened 2018-12-01
base_currency_iso_4217 data errors


levkar:
Some duplicate data in res_currency_data.xml (for example base.EUR)

Some have different names with same id (for example base.TRY) Turkish Lira & New Turkish Lira. 

Shouldn't it be based on https://www.iso.org/iso-4217-currency-codes.html ?

Very nice and necessary  module indeed! Thanks!

@jbeficent

JordiBForgeFlow:
Thanks Levent,

Can you propose a PR?

levkar:
I will do it once i print, sign the CLA and send it to OCA.

pedrobaeza:
@levkar you can sign it digitally if you prefer

levkar:
@pedrobaeza

How? It is not mentioned in OCA page:

https://odoo-community.org/page/cla

pedrobaeza:
You're right: it's not mentioned in the page. I'll update it, but a digital sign on the PDF is also valid.

levkar:
this is fixed

PR #45, opened 2018-12-05
[11.0] iso 4217 Fixed data files, added decimals, python reformat

levkar:
Fixed data files, added decimals, python reformat

### DIFF
--- a/base_currency_iso_4217/README.rst
+++ b/base_currency_iso_4217/README.rst
@@ -32,8 +32,10 @@ ISO 4127 standard:
 
 * Full Name
 
-The module also updates the existing currencies in Odoo,
-based on https://github.com/datasets/currency-codes
+The module also updates the numeric code, full name and decimal places fields
+of existing currencies in Odoo according to ISO 4217:2015.
+
+https://www.iso.org/iso-4217-currency-codes.html
 
 **Table of contents**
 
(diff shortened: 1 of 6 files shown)


levkar:
closes #44 

Some new currencies were not in odoo base (they are not added) and some historic currencies were still there (they weren't updated).

@jbeficent

pedrobaeza:
Why the diff on the XML is so big?

levkar:
I rebuilt the whole XML. No spaces between records would be the reason. Too many changes anyway.

pedrobaeza:
It would be good to let that blank lines to reduce the diff at maximum and review only the real changes. Is it possible?

levkar:
I put spaces but diff is still bad. It should be tabs. Are we supposed to use spaces or tabs in XMLs?

pedrobaeza:
Ouch, I see...

Indeed, the agreement is to use spaces, so we don't gain anything as we shouldn't let that incorrect indentation. @jbeficent please use always spaces as indentation. There's a warning about that on Travis, but not blocking right now.

Then if you prefer, you can remove again the empty line, as I thought it was the cause for the large diff. Sorry for the noise.

levkar:
We can keep it, no problem. I am used to hard OCA accept conditions :+1: 

Once accepted I will send 12.0 version.

pedrobaeza (approved):

PR #46, opened 2018-12-05
[MIG 12.0] migration of product_uom_unece to 12.0

levkar:
just bumped the version.


levkar:
forget to include history. will make another pr

PR #47, opened 2018-12-06
[MIG 12.0] mig unece modules

levkar:
Depends on:

- [x] account_payment_mode - OCA/bank-payment#509

Changed product.uom to uom.uom. Moved test to account_tax_unece and removed the dependency from account_payment_unece.


levkar:
We can also also account_payment_mode dependency from account_payment_unece (it needs payment method form which is somehow in this module) but then we need to strip account.payment.method menu and forms into another module. Let me know if it makes sense.

levkar:
This PR is waiting for https://github.com/OCA/bank-payment/pull/509

pedrobaeza:
@levkar it's better to put on the main comment the dependency for faster reference. I'm adding it right now. Next week I'll push for merging the dependency :smiley:

levkar:
Thanks.

pedrobaeza:
Hi @levkar, I have finished the migration of account_payment_mode. Can you please split the PR having one PR per module?

levkar:
cancelled & created seperate pull requests

PR #53, opened 2018-12-28
[12.0][MIG] - migration l10n_eu_nace

sbejaoui:


pedrobaeza:
Please check Travis.

pedrobaeza:
@sbejaoui please start from #39 instead of 10.0 branch, as this PR contains a good refactoring of the module isolating the NACE tags from partner tags.

sbejaoui:
@pedrobaeza , 

Travis fail on base_iso3166 tests. Can you help me fix it?

The #39 remove data files which essential for this module [l10n_fr_naf_ape](https://github.com/OCA/l10n-france/pull/169). As the PR is not finished yet, and does not take compatibility with other modules, I was thinking that I can base on 10.0 and forward-port the improvements later.

alexis-via (approved):
Just 2 small remarks : replace "OpenERP" by "Odoo" in the rst files and remove the "u" (for unicode) in __manifest__.py now that we use python 3.

remi-filament (approved):

pedrobaeza:
Is this starting from #39 as commented?

legalsylvain:
@pedrobaeza, your question has been answered 8 monthes ago : #39 removes all the data and it seems to be a problem for @sbejaoui.
https://github.com/OCA/community-data-files/commit/1816eeb5e196760df1d297898a16966aa99a3fe6

What do you think about that topic ?
Cc @alexis-via

pedrobaeza:
Well, I still think this needs that refactoring before getting 12.0 branch, or we won't be able to change the structure in stable.

legalsylvain:
Well, I don't know this module at all, so I'm abstain regarding the code. but What i see is a PR for V12 that seems finished and have 2 approvals, and in the meatime a WIP PR that doesn't change since 8 monthes, and has to be fixed, regarding data.
Not sure that it's a good idea to prevent merging this PR.
In the meantime, other PR are waiting this one, like https://github.com/OCA/l10n-france/pull/169

@sbejaoui : could you check why travis is red ? 

regards.

remi-filament:
@legalsylvain Travis failed because of an issue with base_iso3166, which is apparently solved since PR #31 merge
I restarted the build of Travis and it ended OK that time.

sbejaoui:
Hi guys, 
Can we merge this one please?
Kind regards.

pedrobaeza:
As said, I prefer the switch of approach, having a dedicated model. I hadn't time for doing it myself and change l10n_fr_ape for not having excuses for not doing it and I plan to do it next week, but if you want to do it, it's welcome.

sbejaoui:
> As said, I prefer the switch of approach, having a dedicated model. I hadn't time for doing it myself and change l10n_fr_ape for not having excuses for not doing it and I plan to do it next week, but if you want to do it, it's welcome.

Hi Pedro,
We have this need for one of our customers and my colleague @qgroulard will start work on it ASAP. I prefer that we don't mix this refactoring in the migrations PR.

In the same context we need to merge [l10n-france\#169](https://github.com/OCA/l10n-france/pull/169).

pedrobaeza:
@sbejaoui but that's the problem: we can't merge the migration, or we will be slaved in this version of that approach / data structure, as the excuse of having other OCA modules depending on previous data structure will raise, so we need to do the refactoring at the same time than the migration.

sbejaoui:
> @sbejaoui but that's the problem: we can't merge the migration, or we will be slaved in this version of that approach / data structure, as the excuse of having other OCA modules depending on previous data structure will raise, so we need to do the refactoring at the same time than the migration.

make sense, we will do that.

sbejaoui:
Hi @pedrobaeza ,
I pushed a commit for the separation. Can you check please.
Thanks.

njeudy (approved):

yajo (approved):
@pedrobaeza can you review please?

marylla:
@sbejaoui @pedrobaeza Hi there, the PR seems to be ready for a while now. Can we merge?

l10n_eu_nace/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    cr.execute(
+        """
+        UPDATE ir_model_data
+        SET name=concat('old_', name)
```
yajo: Without this patch, possibly this data would be deleted as the module gets updated, and the post migration script would fail:

```suggestion
        SET name=concat('old_', name), noupdate=1
```

l10n_eu_nace/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,13 @@
...
+        WHERE name LIKE 'nace_%'
+        AND module='l10n_eu_nace'
```
yajo: Just to be sure:

```suggestion
        WHERE model = 'res.partner.category'
        AND module='l10n_eu_nace'
```

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+            category = partner.category_id.filtered(
+                lambda c: 'nace_' in c.get_external_id().get(c.id)
+            )
+            if not category:
+                continue
+            if len(category) > 1:
+                category = category[0]
+            partner.nace_id = env.ref(
+                category.get_external_id().get(category.id).replace('old_', '')
+            )
```
yajo: This is possibly a bottleneck if you have lots of partners. Try this:

```suggestion
# Copyright 2019 ACSONE SA/NV
# Copyright 2020 Tecnativa - Jairo Llopis
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).

from openupgradelib import openupgrade

from odoo import SUPERUSER_ID, api


@openupgrade.migrate()
def migrate(env):
    nace_categories = env['ir.model_data'].search([
        ('module', '=', 'l10_eu_nace'),
        ('model', '=', 'res.partner.category'),
    ])
    for category_imd in nace_categories:
        partners = env['res.partner'].search([
            ('category_id', 'in', [category_imd.res_id]),
        )
        partners.write({
            'nace_id': env.ref('l10n_eu_nace.%s' % category_imd.name.replace('old_', '', 1)).id,
        })
```

Benefits:

1. Single search for all categories and their XMLID names.
1. Write all partners in a category in batch.
1. More readable.

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+
+
+class ResPartnerNace(models.Model):
+
+    _name = 'res.partner.nace'
+    _description = 'European NACE partner categories'
```
yajo:
```suggestion
    _description = 'European NACE partner category'
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
...
+    _order = 'code'
+    _parent_store = True
+
+    name = fields.Char()
+    parent_id = fields.Many2one(comodel_name="res.partner.nace")
+    code = fields.Char()
+    child_ids = fields.One2many(
+        'res.partner.nace', 'parent_id', string='NACE subcategories'
+    )
+    active = fields.Boolean(default=True)
```
yajo: This is used by default to filter:

```suggestion
    active = fields.Boolean(index=True, default=True)
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
...
+class ResPartnerNace(models.Model):
+
+    _name = 'res.partner.nace'
+    _description = 'European NACE partner categories'
+    _order = 'code'
+    _parent_store = True
+
+    name = fields.Char()
+    parent_id = fields.Many2one(comodel_name="res.partner.nace")
+    code = fields.Char()
```
yajo: This is used as default order:

```suggestion
    code = fields.Char(index=True)
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ResPartnerNace(models.Model):
+
+    _name = 'res.partner.nace'
+    _description = 'European NACE partner categories'
+    _order = 'code'
+    _parent_store = True
+
+    name = fields.Char()
```
yajo: This will be used to search. Also it should be translatable, right?:

```suggestion
    name = fields.Char(index=True, translate=True)
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
...
+        'res.partner.nace', 'parent_id', string='NACE subcategories'
+    )
+    active = fields.Boolean(default=True)
+    parent_path = fields.Char(index=True)
+    partner_ids = fields.One2many(
+        comodel_name="res.partner", inverse_name="nace_id", string="Partners"
+    )
+
+    @api.multi
+    def name_get(self):
```
yajo: I'm more in favor of declaring a computed `display_name` field, for readability, but OK. :wink:

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
...
+
+class ResPartnerNace(models.Model):
+
+    _name = 'res.partner.nace'
+    _description = 'European NACE partner categories'
+    _order = 'code'
+    _parent_store = True
+
+    name = fields.Char()
+    parent_id = fields.Many2one(comodel_name="res.partner.nace")
```
yajo: m2o always should be indexed: https://github.com/odoo/odoo/pull/39332

```suggestion
    parent_id = fields.Many2one(comodel_name="res.partner.nace", index=True)
```

l10n_eu_nace/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+
+    _inherit = 'res.partner'
+
+    nace_id = fields.Many2one(comodel_name="res.partner.nace", string="NACE")
```
yajo: Always index these: https://github.com/odoo/odoo/pull/39332

```suggestion
    nace_id = fields.Many2one(comodel_name="res.partner.nace", string="NACE", index=True)
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,54 @@
...
+                        names.append(current.name)
+                    current = current.parent_id
+                res.append((category.id, ' / '.join(reversed(names))))
+        return res
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        return self.search(
+            ['|', ('name', operator, name), ('code', 'ilike', name)]
+        ).name_get()
```
yajo: :no_entry_sign: This is wrong, sorry. Try this:

```suggestion
    def name_search(self, name, args=None, operator='ilike', limit=100):
        args = list(args or [])
        args += ['|', ('name', operator, name), ('code', 'ilike', name)]
        return super().name_search(name, args, operator, limit)
```

l10n_eu_nace/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* Lionel Sausin (Numérigraphe) <ls@numerigraphe.com>
+* Sistheo
+* data files courtesy of the European Union
+* Souheil Bejaoui <souheil.bejaoui@acsone.eu>
```
yajo, line 4: I wrote enough code in this review :sweat_smile: 

```suggestion
* Souheil Bejaoui <souheil.bejaoui@acsone.eu>
* Jairo Llopis <jairo.llopis@tecnativa.com>
```

l10n_eu_nace/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="res_partner_form_view">
```
yajo: I don't see the need to modify this XMLID:

```suggestion
    <record model="ir.ui.view" id="view_partner_form">
```

yajo (changes requested):
Some blocking points.

@Tecnativa TT17678

sbejaoui:
Hi @Yajo ,

I applied all you suggestions, can you update your review.

Thanks.

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -45,9 +45,9 @@ def name_get(self):
 
     @api.model
     def name_search(self, name, args=None, operator='ilike', limit=100):
-        return self.search(
-            ['|', ('name', operator, name), ('code', 'ilike', name)]
-        ).name_get()
+        args = list(args or [])
+        args += ['|', ('name', operator, name), ('code', 'ilike', name)]
```
yajo: Oops, I forgot this nasty one:

```suggestion
        args = ['|', ('name', operator, name), ('code', 'ilike', name)] + args
```

yajo (changes requested):
@pedrobaeza Please could you confirm us these migration scripts are gonna work?

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -46,7 +46,7 @@ def name_get(self):
     @api.model
     def name_search(self, name, args=None, operator='ilike', limit=100):
         args = list(args or [])
-        args += ['|', ('name', operator, name), ('code', 'ilike', name)]
+        args = ['|', ('code', 'ilike', name)] + args
```
yajo: This is wrong, sorry:

```suggestion
        args = ['|', ('name', operator, name), ('code', 'ilike', name)] + args
```

yajo (changes requested):

yajo (approved):
There's [a failure in Travis](https://travis-ci.org/github/OCA/community-data-files/jobs/696406353#L1026-L1040), please check it.

sbejaoui:
> There's [a failure in Travis](https://travis-ci.org/github/OCA/community-data-files/jobs/696406353#L1026-L1040), please check it.

Fixed and squashed

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,56 @@
...
+                    current = current.parent_id
+                res.append((category.id, ' / '.join(reversed(names))))
+        return res
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        args = list(args or [])
+        if not (name == "" and operator == "ilike"):
+            args += ["|"]
+        args += [("code", operator, name)]
```
yajo: Oh please, not again! :grin: 

Look, I think you're not understanding my previous comments on this part. You should **always** put trusted parts of the domain **before** untrusted ones. It's a safety measure. Actually this code is buggy and if you look close at it you'll see it.

Check this talk on the matter: https://youtu.be/27K-LmTPnZQ?t=2965

You can also use `odoo.osv.expression.OR/AND` if you don't want to worry about the details.

yajo (changes requested):

l10n_eu_nace/models/res_partner_nace.py
sbejaoui: It's now when I get your point :pensive: .

I didn't think at all about that as it's already done in super (args +=) :confused:, 

https://github.com/odoo/odoo/blob/c881e3fe2a8b3db43fa9613a52b996fdfab21392/odoo/models.py#L1663-L1676

sbejaoui: I pushed a new commit (even it has no effect)

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+def migrate(env):
+    nace_categories = env['ir.model_data'].search(
+        [
+            ('module', '=', 'l10_eu_nace'),
+            ('model', '=', 'res.partner.category'),
+        ]
+    )
+    for category_imd in nace_categories:
+        partners = env['res.partner'].search(
+            [('category_id', 'in', [category_imd.res_id])]
```
pedrobaeza: Shouldn't you look only for records with `old` prefix?

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+        )
+
+        partners.write(
+            {
+                'nace_id': env.ref(
```
pedrobaeza: Better to do this in SQL and in one shot (more or less this query - not exact):
```
UPDATE res_partner rp
SET nace_id = imd2.res_id
FROM ir_model_data imd1
LEFT JOIN ir_model_data imd2
    ON  imd2.module = imd1.module
    AND imd2.name = substring(imd1.name, 4)
WHERE imd1.name LIKE 'old_%'
    AND rp.category_id = imd1.res_is
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,57 @@
...
+                    current = current.parent_id
+                res.append((category.id, ' / '.join(reversed(names))))
+        return res
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        args = list(args or [])
+        domain = AND([[("code", operator, name)], args])
+        if not (name == "" and operator == "ilike"):
+            domain = ["|"] + domain
```
yajo: This code is buggy. Please remove it, or explain to me what you're trying to do so I can help you fix it :wink: 

```suggestion
```

yajo (changes requested):

l10n_eu_nace/models/res_partner_nace.py
sbejaoui: Please check to understand it, https://github.com/odoo/odoo/blob/c881e3fe2a8b3db43fa9613a52b996fdfab21392/odoo/models.py#L1663-L1676

in super, odoo add to args `[(self._rec_name, operator, name)]`. The OR here is needed to end with `['|', ('code', operator, name), (self._rec_name, operator, name)]`

you removed it in your last commit and break the name search.

yajo: But that's an unsafe way to do it, because if you get some `args`, the `|` part will only work for the 1st 2 leafs.

If you really want to avoid that part, what you have to do is ... (see below)

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,57 @@
...
+        return super().name_search(name, domain, operator, limit)
```
yajo: ... this:

```suggestion
        return super().name_search(name or self._rec_name, domain, operator, limit)
```

And define a `_rec_name` in the model (there's none AFAICS).

rafaelbn:
Please @sbejaoui , could we finish this? We are in a hurry

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
sbejaoui: I did it as you suggested but couldn't test it as all our db are migrated. It will be nice if you can run it on existing data.

l10n_eu_nace/models/res_partner_nace.py
sbejaoui: I don't get you, 

argument name contain the search value, in case it's empty it will search for nace with name "display_name"?

IMO, there is no way to override name_search without changing the domain in unsafe way due to _name_search implementation in base model.

Avoiding the security leak has no added value as the leak is in base model

This explain why many of OCA and Odoo modules when overriding name_search don't send back to super and return the result of normal search and for the same reasons you last fix suggestions broke the tests.

So, I decide to give up on name_search and use a computed with a dedicated search method

sbejaoui:
> Please @sbejaoui , could we finish this? We are in a hurry

sorry for the late, I believe it's all good now

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+def migrate(cr, version):
+    cr.execute(
+        """
+            UPDATE res_partner rp
+            SET nace_id = imd2.res_id
+            FROM ir_model_data imd1
+            LEFT JOIN ir_model_data imd2
+                ON  imd2.module = imd1.module
+                AND imd2.name = substring(imd1.name, 4)
```
yajo: Oops, a little big bug here:

```suggestion
                AND imd2.name = substring(imd1.name, 5)
```

FTR:

```sql
devel=# select substring('old_hi', 4);
 substring 
-----------
 _hi
(1 row)

devel=# select substring('old_hi', 5);
 substring 
-----------
 hi
(1 row)
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,65 @@
...
+                ('name', operator, value),
+                ('code', operator, value),
+            ]
+        else:
+            domain = [
+                '|',
+                ('name', operator, value),
+                ('code', operator, value),
+            ]
+        return [('id', 'in', self.search(domain).ids)]
```
yajo: Oops there's something weird.

```suggestion
        return domain
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,65 @@
...
+
+    _sql_constraints = [
+        ("ref_code", "unique (code)", _("NACE Code must be unique!"))
```
yajo: Translations are applied later by Odoo.

```suggestion
        ("ref_code", "unique (code)", "NACE Code must be unique!")
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,65 @@
...
+    active = fields.Boolean(index=True, default=True)
+    parent_path = fields.Char(index=True)
+    partner_ids = fields.One2many(
+        comodel_name="res.partner", inverse_name="nace_id", string="Partners"
+    )
+
+    @api.depends("code", "name")
+    def _compute_complete_name(self):
+        for category in self:
+            if self._context.get('nace_display') == 'short':
```
yajo: Don't you think this should be the opposite?

I mean: the short method is quick, while the long method is possibly a bottleneck (recurses over parents and then reverses a list for each record, both expensive operations).

So, I think that the short method should be used unless explicitly asked by `nace_display=long` context. This way we don't penalize performance unless needed.

What do you think?

BTW, this solution of a computed searchable field is way better than the previous one, good idea!

yajo (changes requested):
This looks much better, but the new code has its new bugs.

Be patient, we're getting close! Thanks.

l10n_eu_nace/models/res_partner_nace.py
sbejaoui: I agree with you that's better to use `nace_display=long` but it can break inheritance if it's used in some custom modules as this option was there since previous versions.

yajo: It's not a problem. We are migrating the module, it's the perfect timing to do breaking changes. :wink:

sbejaoui: done

yajo (approved):

l10n_eu_nace/i18n/ru.po
```diff
@@ -0,0 +1,5236 @@
+# Translation of OpenERP Server.
```
yajo: This file is broken: https://travis-ci.org/github/OCA/community-data-files/jobs/701976078#L1278-L3835

Please delete it for the bot to be able to continue with the merge.

l10n_eu_nace/i18n/hr.po
```diff
@@ -0,0 +1,5003 @@
+# Translation of OpenERP Server.
```
yajo: This file is broken: https://travis-ci.org/github/OCA/community-data-files/jobs/701976078#L3837-L3866

Please delete it so the bot can continue with the merge.

yajo (changes requested):

l10n_eu_nace/i18n/hr.po
sbejaoui: done

l10n_eu_nace/data/make_data.py
```diff
@@ -0,0 +1,106 @@
...
+"Language-Team: \\n"
+"MIME-Version: 1.0\\n"
+"Content-Type: text/plain; charset=UTF-8\\n"
+"Content-Transfer-Encoding: \\n"
+"Plural-Forms: \\n"
+
+"""
+    )
+    for row in src:
+        name = "[%s] %s" % (row[1], row[2])
```
yajo, line 95: FTR those failing translations are probably caused by this line, which should escape properly special characters such as `\n` or `"`.

yajo (approved):

PR #54, opened 2019-01-09
[11.0][MIG] product_uom > product_uom_extra_data

MiquelRForgeFlow:
Rename and repository change suggested in https://github.com/OCA/product-attribute/pull/423#issuecomment-452432077.


product_uom_extra_data/migrations/11.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2018 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+xmlid_renames = [
```
pedrobaeza: No need of this. Simply add the renaming of the module in apriori file in OpenUpgrade.

pedrobaeza (changes requested):

product_uom_extra_data/migrations/11.0.1.0.0/pre-migration.py
MiquelRForgeFlow: Good

MiquelRForgeFlow:
Ready

pedrobaeza (approved):

HviorForgeFlow (approved):

PR #59, opened 2019-01-24
[11.0] [MIG] l10n_eu_nace

remihb:
Standard migration from 10.0.
See here for improved migration to 11.0 : [https://github.com/OCA/community-data-files/pull/60](https://github.com/OCA/community-data-files/pull/60)

There's still lint errors because of CSV files not used from manifest -- W7930(file-not-used), and there's no dirs that make sense for this files : ['static', 'test', 'tests', 'migrations'] (default excluded dirs that avoid the error)
Maybe reconsider this [https://github.com/OCA/pylint-odoo/pull/217](https://github.com/OCA/pylint-odoo/pull/217)

PR #60, opened 2019-01-24
[IMP] l10n_eu_nace : custom migration to 11.0

remihb:
Custom migration from 10.0
See here for standard migration to 11.0 : [https://github.com/OCA/community-data-files/pull/59](https://github.com/OCA/community-data-files/pull/59)

Changes are :
- The data imported into OpenERP is generated from the files downloaded from the RAMON service - [http://ec.europa.eu/eurostat/ramon/nomenclatures/index.cfm?TargetUrl=LST_CLS_DLD&StrNom=NACE_REV2&StrLanguageCode=FR&StrLayoutCode=#
](http://ec.europa.eu/eurostat/ramon/nomenclatures/index.cfm?TargetUrl=LST_CLS_DLD&StrNom=NACE_REV2&StrLanguageCode=FR&StrLayoutCode=#)
- Automatic translation
- NACE codes have their specific class res_partner_nace instead of putting them into res_partner_category


pedrobaeza:
I don't get why having 2 PRs for migrating a module, and more having #39 that already migrates the module.

remihb:
Ok sorry, I've forgotten that PR [https://github.com/OCA/community-data-files/pull/39](https://github.com/OCA/community-data-files/pull/39)

But this one is more like a proposition, inspired by [https://github.com/Antiun/antiun-odoo-addons/tree/8.0/partner_nace](https://github.com/Antiun/antiun-odoo-addons/tree/8.0/partner_nace)
It's mainly for avoid mixing NACE with other res_partner_category, as suggested in [https://github.com/OCA/community-data-files/pull/39#issuecomment-399187306](https://github.com/OCA/community-data-files/pull/39#issuecomment-399187306)

pedrobaeza:
Yeah, I agree on the approach change, but then the proper way to do that is to preserve commit history and then add on top the commit adding that proposal. Can you please do it and we will close #39  and #59 ?

l10n_eu_nace/migrations/11.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,37 @@
...
+            "SELECT name FROM res_partner_category WHERE id IN "
+            "(SELECT category_id FROM res_partner_res_partner_category_rel "
+            "WHERE partner_id = %s)" % (partner,))
+        categs = env.cr.fetchall()
+        if not categs:
+            continue
+        old_naces = "\n".join([categ[0] for categ in categs])
+        env.cr.execute(
+            "UPDATE res_partner SET old_naces_temp = '%s' WHERE id = %s" % (
+                old_naces, partner))
```
sebalix: Hi Rémi,
As you are performing SQL queries only in this pre-migration script you don't need `env` at all. Just pass the `cr` cursor to `migrate_nace` and it'll be fine. BTW I don't know how Odoo behaves if you use an environment in a pre-migration script... but it seems to work if you tested it ;)

l10n_eu_nace/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    nace_ids = fields.Many2many(
+        'res.partner.nace', string="Activities (NACE)", ondelete='set null')
```
sebalix: `ondelete` useless on  Many2many

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,46 @@
...
+    code = fields.Char(required=True)
+    name = fields.Char(required=True, translate=True)
+    generic = fields.Char(string="ISIC Rev.4")
+    rules = fields.Text()
+    central_content = fields.Text(translate=True, string="Contents")
+    limit_content = fields.Text(translate=True, string="Also contents")
+    exclusions = fields.Char(string="Excludes")
+    parent_id = fields.Many2one(
+        'res.partner.nace', ondelete='restrict')
+    children = fields.One2many('res.partner.nace',
```
sebalix: can you add the suffix `_ids`?

l10n_eu_nace/views/res_partner.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
```
sebalix: `<odoo>`, and you can drop `<data>`

l10n_eu_nace/wizards/nace_import.py
```diff
@@ -0,0 +1,293 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
sebalix: Please use short copyrights

sebalix (changes requested):
Not tested, just a code review, minor things to change.

l10n_eu_nace/migrations/11.0.2.0.0/pre-migrate.py
remihb: I'll see many migration scripts using `@openupgrade.migrate()` as decorator for migrate methods by importing `from openupgradelib import openupgrade`.

As far as I can see, it doesn't seems usefull from Odoo versions above 9.0. So should I use it?

l10n_eu_nace/wizards/nace_import.py
remihb: Sorry for Antiun but I remove their copyrights. Is that okay to do that?

l10n_eu_nace/migrations/11.0.2.0.0/pre-migrate.py
pedrobaeza: @sebalix although the environment is not used right now, it can be useful in the future, so don't dismiss it, as `env.cr` or `cr = env.cr` is very light to be written.

@remihb the decorator serves for 2 purposes:

- Don't execute the migration script if no version is provided, which happens when installing module for the first time, but indeed this is deactivated since a lot (I think in v10).
- Auto creating an environment for you.

so I wouldn't remove it. In the future, we can even attach more logic to this simply updating the openupgradelib library and then all the scripts will get benefit from it without changing anything.

remihb: Okay thx, ready for appproval ;)

l10n_eu_nace/wizards/nace_import.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data>
```
sebalix: `<openerp><data>` to replace by `<odoo>`

remihb: Dayum

sebalix: ... :D

sebalix (approved):
LGTM

l10n_eu_nace/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+import re
+import odoo
+
+from odoo import api, SUPERUSER_ID
+
+logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
```
yajo: I encourage you to use `openupgradelib` in these scripts.

For example, you can write here:

```python
@openupgrade.migrate()
def migrate(env, version):
```

l10n_eu_nace/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,52 @@
...
+                    new_env = odoo.api.Environment(
+                        new_cr, env.uid, env.context)
+                    action(new_env)
+
+
+def import_nace(env):
+    logger.info(
+        "[IMPORT res.partner.nace] download from ramon european service")
+    wiz_model = env['nace.import']
+    vals = wiz_model.default_get(wiz_model._fields)
```
yajo: You don't need to specify the defaults as vals. That's why they are defaults!

l10n_eu_nace/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,52 @@
...
+    wiz = wiz_model.create(vals)
```
yajo: ... so just do:

```suggestion
    wiz = wiz_model.create({})
```

l10n_eu_nace/migrations/11.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,37 @@
...
+
+def migrate(cr, version):
+    if not version:
+        return
+    with api.Environment.manage():
+        migrate_nace(cr)
+
+
+def migrate_nace(cr):
+    logger.info(
```
yajo, line 21: To get a proper name for the column, use [`openupgrade.get_legacy_name("old_nace")`](https://github.com/OCA/openupgradelib/blob/973e13f84087fe2d81a59e78c5699d41d6c7dcbd/openupgradelib/openupgrade.py#L1073-L1083)

l10n_eu_nace/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,52 @@
...
+        "[UDPATE res.partner] reimport old naces from 'old_naces_temp' in "
+        "res_partner to res_partner_nace")
+    env.cr.execute(
+        "SELECT id, old_naces_temp FROM res_partner "
+        "WHERE old_naces_temp IS NOT NULL")
+    partners = env.cr.fetchall()
+    for id_, old_naces in partners:
+        for category in old_naces.split('\n'):
+            if re.match("^\[.*\]", category):
+                nace_code = category.split("[")[1].split("]")[0]
```
yajo: You can use the `re.match()` itself to capture this code. Example:

```python
>>> import re
>>> tag = "[a.b] nace thingie"                                                                                                                                                                   
>>> re.match(r"^\[(.*)\]", tag)[1]                                                                                                                                                               
'a.b'
```

The `1` key is capturing whatever is found inside the parenthesis in the regexp.

l10n_eu_nace/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,52 @@
...
+                nace = env['res.partner.nace'].search([
+                    ('code', '=', nace_code)])
+                if nace:
+                    nace.ensure_one()
+                    partner = env['res.partner'].with_context(
+                        active_test=False).browse(id_)
+                    partner.nace_ids |= nace
```
yajo: If a partner has only 1 NACE, you should set it into `primary_nace_id`. If it has more than 1, set them in `secondary_nace_ids`.

(See below about these columns).

l10n_eu_nace/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019 ABF OSIELL <http://osiell.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    nace_ids = fields.Many2many(
```
yajo: Just like [`partner_nace` had](https://github.com/Antiun/antiun-odoo-addons/blob/bcc0b3c4762b5a5e04dfbe952a15981ad19c78ee/partner_nace/models/res_partner.py#L31-L35):

```suggestion
    primary_nace_id = fields.Many2one(
        "res.partner.nace",
        string="Primary NACE Activity",
    )
    secondary_nace_ids = fields.Many2many(
```

l10n_eu_nace/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
...
+        'res.partner.nace', string="Activities (NACE)")
```
yajo:
```suggestion
        'res.partner.nace', string="Secondary NACE Activities")
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,27 @@
...
+    level = fields.Integer(required=True)
+    code = fields.Char(required=True)
+    name = fields.Char(required=True, translate=True)
+    generic = fields.Char(string="ISIC Rev.4")
+    rules = fields.Text()
+    central_content = fields.Text(translate=True, string="Contents")
+    limit_content = fields.Text(translate=True, string="Also contents")
+    exclusions = fields.Char(string="Excludes")
+    parent_id = fields.Many2one(
+        'res.partner.nace', ondelete='restrict')
```
yajo:
```suggestion
        'res.partner.nace', string="Parent code", ondelete='restrict')
```

l10n_eu_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,27 @@
...
+    child_ids = fields.One2many(
+        'res.partner.nace', inverse_name='parent_id')
```
yajo:
```suggestion
        'res.partner.nace', string="Child codes", inverse_name='parent_id')
```

l10n_eu_nace/readme/INSTALL.rst
```diff
@@ -0,0 +1,3 @@
+To install this module, you need request python module:
+
+* pip intall requests
```
yajo: :fire: Delete this file.

This requirement [is already a dependency in Odoo 10](https://github.com/odoo/odoo/blob/25155aa47244aa6e8216598be94a4acdfe1f1c07/requirements.txt#L33). It will always be present.

l10n_eu_nace/wizards/nace_import.py
```diff
@@ -0,0 +1,274 @@
...
+    _url_path = '/eurostat/ramon/nomenclatures/index.cfm'
+    _url_params = {
+        'TargetUrl': 'ACT_OTH_CLS_DLD',
+        'StrNom': 'NACE_REV2',
+        'StrFormat': 'XML',
+        'StrLanguageCode': 'EN',
+        # 'IntKey': '',
+        # 'IntLevel': '',
+        # 'TxtDelimiter': ';',
+        # 'bExport': '',
```
yajo: Remove these comments.

l10n_eu_nace/wizards/nace_import.py
```diff
@@ -0,0 +1,274 @@
...
+        return item
+
+    def _download_nace(self, lang_code):
+        params = self._url_params.copy()
+        params['StrLanguageCode'] = lang_code
+        url = self._url_base + self._url_path + '?'
+        url += '&'.join([k + '=' + v for k, v in params.items()])
+        logger.info('Starting to download %s', url)
+        try:
+            res_request = requests.get(url)
```
yajo: Add a timeout.

l10n_eu_nace/wizards/nace_import.py
```diff
@@ -0,0 +1,274 @@
...
+                         ('lang', '=', lang.code)]
+                translation = translation_model.search(query)
+                data = {
+                    'value': value,
+                    'state': 'translated'
+                }
+                if translation:
+                    translation_model.write(data)
+                    # TODO  2019-01-18: dangerous code, write in ir_translation
+                    self.env.cr.execute(
```
yajo: As the comment says, this is dangerous. Just do something similar to:

```py
nace_record.with_context(lang=target_lang).write(translated_vals)
```

l10n_eu_nace/wizards/nace_import.py
```diff
@@ -0,0 +1,274 @@
...
+        # Available lang list
+        langs = lang_model.search(
+            [('code', 'in', list(self._available_langs.keys())),
+             ('active', '=', True)])
+        # All current NACEs, delete if not found above
+        naces_to_delete = nace_model.search([])
+        # Download NACEs in english, create or update
+        logger.info('Import NACE Rev.2 English')
+        xmlcontent = self._download_nace('EN')
+        # pylint: disable=c-extension-no-member
```
yajo:
```suggestion
```

l10n_eu_nace/wizards/nace_import.py
```diff
@@ -0,0 +1,274 @@
...
+            xmlcontent = self._download_nace(nace_lang)
+            dom = etree.fromstring(
+                xmlcontent.encode('utf-8'), parser=utf8_parser)
+            for node in dom.iter('Item'):
+                logger.debug(
+                    'Reading lang=%s, level=%s, code=%s', nace_lang,
+                    node.get('idLevel', 'N/A'), node.get('id', 'N/A'))
+                self.translate_nace(node, lang)
+        # Delete obsolete NACEs
+        if naces_to_delete:
```
yajo: The wizard should have a bool option for this:

```suggestion
        if self.delete_obsolete_records and naces_to_delete:
```

l10n_eu_nace/views/res_partner_nace.xml
```diff
@@ -0,0 +1,57 @@
...
+                    <field name="rules"/>
+                    <field name="generic"/>
+                    <field name="central_content"/>
+                    <field name="limit_content"/>
+                    <field name="exclusions"/>
+                </group>
+            </form>
+        </field>
+    </record>
+
```
yajo, line 52: Please add a search view also, to allow quick searches based on naces.

yajo (changes requested):
This is indeed a great improvement.

Some things need to be fixed/improved though. See comments below.

Thanks a lot!

PR #61, opened 2019-01-26
[FIX][12.0][base_iso3166 ] make historic countries work

joshuajan:
Hello, This PR will fix this problem:

When the country code is not exist. This code will return False. 
`c = pycountry.countries.get(alpha_2=country.code)`

So the `c = pycountry.historic_countries.get(alpha_2=country.code)` will never be call. 

And it make the Test failed: E.g: https://travis-ci.org/OCA/community-data-files/jobs/479310840

### DIFF
--- a/base_iso3166/README.rst
+++ b/base_iso3166/README.rst
@@ -79,6 +79,7 @@ Contributors
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
 * Enric Tobella <etobella@creublanca.es>
 * Aitor Bouzas <aitor.bouzas@adaptivecity.com>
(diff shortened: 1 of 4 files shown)


pedrobaeza:
The code you are putting works with latest version of the library, but you should preserve the compatibility with all code versions where current code worked, so an extra step should be done. It's a really annoyance that this people modifies so much the library, but we must adapt to all possibilities.

joshuajan:
Hello @pedrobaeza 

Thanks for your reply. I have modified the code. It work with pycountry >= 0.19 and pycountry >= 16.11.8

pedrobaeza (approved):
We can fast-track this for solving the CI status.

PR #63, opened 2019-03-04
[11.0][IMP] Add the number of digits according the ISO 4217

sylvainvh:
I'm not using the field rounding because I want to make the different between the number of digits (according ISO 4217) and how the user want to display this currency


pedrobaeza:
Isn't the field better named `digits_number`?

sylvainvh:
Hi @pedrobaeza , you right ! I will change the field name

sylvainvh:
Done !

pedrobaeza (approved):

sylvainvh:
Someone to validate my PR ?

pedrobaeza:
Let's fast track this as straight-forward

PR #64, opened 2019-03-27
IMP show acc_type on bank account list on partners

angelmoya:
In version 12, you can create bank account directly on bank account list on partner form.

See account type directly on this list is usefull to check if bank account is iban or not.

### DIFF
--- a/base_bank_from_iban/README.rst
+++ b/base_bank_from_iban/README.rst
@@ -23,7 +23,7 @@ Bank from IBAN
     :target: https://runbot.odoo-community.org/runbot/101/12.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module adds a code to bank definition for using it as matching for filling the bank
 from the IBAN bank account number. It uses the existing by country bank mapping in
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why it should be this here? There's a risk of other modules doing the same and have duplicated field. Have you seen if bank-payment modules do the same? And any way, not sure this field is interesting to be shown.  It's very easy to group by that field, and more and more, all accounts are IBAN.

angelmoya:
Sure, all accounts are IBAN, this way you can now if the bank account that you are writting is right.

If you are creating new account for a customer, and bank account is not right then you will see that account type is not IBAN... then you now that there are a mistake... it could be use as IBAN validation.

I think this is usefull for all modules that use IBAN banks, and is better to habe here instead in other modules.

pedrobaeza:
Can you put an screenshot of where exactly this is shown?

angelmoya:
This is an screenshot of this runbot

Azure Interior   Odoo

pedrobaeza:
Right, where I was suspecting... If I don't remember bad, on the bank account form, when you create it, you already have that field, so you can see there if it's correct or not, so I still think that this is not very good to be added, and more in this module.

What do you think @sbidoul @carlosdauden ?

angelmoya:
OK, we wait for other reviews.

I know that in bank account form you get this field, but this bank account list on partner form is new in version 12.

I think that if users have the posibility to create bank accounts directly on this partner form, they will use.

I think that this is an usability improvement.

carlosdauden (approved):
LGTM

sbidoul:
The usability improvement might be useful, but I think it's not the right module to add it.

The problem is there is no good place for such usability improvement modules, or we would end up adding one module for each small improvement. For this @alexis-via 's approach to usability modules is interesting. Perhaps we should consider adopting such modules in OCA.

pedrobaeza:
The problem with that modules is that are very opinionated, and if I want to add one field to that module, other contributor (or the original one) might not want it, and we will end up with a locking situation.

sbidoul:
I agree consensus would be hard to reach on such modules. Might be worth trying anyway.

PR #66, opened 2019-05-24
[12.0][MIG]  account_payment_unece: add migration scripts

MiquelRForgeFlow:
Add migration scripts for `account_payment_unece`.

### DIFF
--- a/account_payment_unece/__manifest__.py
+++ b/account_payment_unece/__manifest__.py
@@ -9,7 +9,7 @@
     'license': 'AGPL-3',
     'summary': 'UNECE nomenclature for the payment methods',
     'author': 'Akretion,Odoo Community Association (OCA)',
-    'website': 'http://www.akretion.com',
+    'website': 'https://github.com/OCA/community-data-files',
     'depends': [
         'account_payment_mode',
         'base_unece',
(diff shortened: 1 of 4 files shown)


pedrobaeza:
But this is not something about migration, but maybe incorrect module definition. Are original XML-IDs noupdate? If not, although this module put noupdate="1", they will remain as noupdate="0". If yes, then this module must modify them through post_init_hook, not with XML data.

MiquelRForgeFlow:
Original xmlids in `account` are also noupdate="1".

pedrobaeza:
Then installing this module doesn't work at all with that payment modes. As said, this is not migration script, but post_init_hook. If previous version is the same, you would need a migration script calling post_init_hook.

MiquelRForgeFlow:
what do you mean with "if previous version is the same"? In v11 they were ref=payment_means_31 and in this v12 they are ref=payment_means_30.

pedrobaeza:
Then none of them versions are going to work installing from scratch. That data is not going to be populated when installing the module. You need to put the post_init_hook and in this 12.0 version, add this migration script, but calling the post_init_hook.

MiquelRForgeFlow:
ok

MiquelRForgeFlow:
@pedrobaeza is it ok, now? 🤔

pedrobaeza:
The same comment about noupdate records in init mode, so please fix this and use load noupdate records in migration scripts.

MiquelRForgeFlow:
done

pedrobaeza (approved):

PR #67, opened 2019-05-24
[11.0][FIX] account_payment_unece: add post_init_hook

MiquelRForgeFlow:
The records in `account_payment_method.xml` were not running because the the original xmlids in `account` module were already noupdate.


MiquelRForgeFlow:
@pedrobaeza

pedrobaeza:
We discover recently that in init (install) mode, noupdate records are updated any way, so this shouldn't be needed, so closing.

PR #69, opened 2019-09-23
[10.0] country.code_numeric = c.numeric, 'NoneType' object has no attribute 'numeric'

JoanMarin:
I have Odoo installed (OCB version 10.0), on ubuntu server 18.04 LTS, when we install the base_iso3166 module, the following error occurs:

OCA/community-data-files/base_iso3166/models/res_country.py", line 37, in _compute_codes
    country.code_numeric = c.numeric
AttributeError: 'NoneType' object has no attribute 'numeric'

To solve this error, it worked for me to make the change that I propose in this PR.

### DIFF
--- a/base_iso3166/models/res_country.py
+++ b/base_iso3166/models/res_country.py
@@ -32,9 +32,15 @@ class ResCountry(models.Model):
                     c = pycountry.countries.get(alpha_2=country.code)
                 except KeyError:
                     c = pycountry.countries.get(alpha2=country.code)
-                country.code_alpha3 = getattr(c, 'alpha_3',
-                                              getattr(c, 'alpha3', False))
-                country.code_numeric = c.numeric
+
+                country.code_alpha3 = getattr(
+                    c,
+                    'alpha_3',
+                    getattr(c, 'alpha3', False))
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Have you updated pycountry to supported versions?

JoanMarin:
Yes, I installed the python dependencies, as indicated in the requirements.txt file of the repository::

https://github.com/OCA/community-data-files/blob/10.0/requirements.txt

And I got the latest version of pycountry (19.8.18):

https://pypi.org/project/pycountry/

image

pedrobaeza:
The problem seems to be with latest versions. Try exactly with the version 0.19 indicated.

JoanMarin:
I did the test with pycountry == 0.19, and the module installation worked correctly.

You would have to make the change in the requirements.txt file to set pycountry == 0.19 and for avoid future problems or how would you proceed?

pedrobaeza:
There are 2 ways:

1. Pin that version in the requirements file.
2. Update the module for working with all the versions (not the change you are doing, as you are breaking then those with that old version installed in the system).

It's really annoying that this library changes its API with each version, but we don't have better option...

PR #86, opened 2020-03-04
[13.0] [ADD] Product Meat UNECE

ThomasBinsfeld:


chrisdec (approved):
functional test

product_meat_unece/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Product Meat Unece",
+    "summary": """
+        This module adds the UNECE Meat Carcasses and Cuts Classification.""",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://acsone.eu/",
```
acsonefho: I think the website should be the one of the OCA repo

product_meat_unece/models/product_template.py
```diff
@@ -0,0 +1,170 @@
...
+        "unece_meat_refrigeration_id",
+        "unece_meat_bovine_category_id",
+        "unece_meat_production_system_id",
+        "unece_meat_feeding_system_id",
+        "unece_meat_post_slaughter_system_id",
+        "unece_meat_fat_thickness_id",
+        "unece_meat_bovine_quality_system_id",
+        "unece_meat_weight_range_id",
+        "unece_meat_packing_id",
+        "unece_meat_conformity_assessment_id",
```
acsonefho: To have a perfect depends, you can use a lambda function into the `@api.depends()`, the function only have to return a `list` of fields (`str`).
A great improvement could be to return every fields with `unece_meat_code = True`.
So new modules with `unece_meat_code` will be automatically added.

product_meat_unece/models/unece_code_list.py
```diff
@@ -0,0 +1,67 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class UneceCodeList(models.Model):
+    _inherit = "unece.code.list"
+
+    type = fields.Selection(
```
acsonefho, line 11: IMO a field named `type` is not correct because `type` is a python's reserved word.

acsonefho (commented):
Only minors changes (but no blocking points).
Tested functionally and it works perfectly.

product_meat_unece/models/unece_code_list.py
ThomasBinsfeld: I agree, but that's defined in `base_uence`, which is merged in `13.0` already. That can be refactored in a separated PR.

product_meat_unece/__manifest__.py
ThomasBinsfeld: Fixed

product_meat_unece/models/product_template.py
ThomasBinsfeld: Good idea, done

acsonefho (approved):

PR #88, opened 2020-03-11
[13.0] [IMP] Base UNECE: allow search on code

ThomasBinsfeld:
Allow search on code (if code matched exactly), fallback on regular name search otherwise.


astirpe (approved):
LGTM

PR #90, opened 2020-06-25
[IMP] l10n_eu_nace: add secondary

yajo:
Some need to add secondary NACE codes to partners.

This is a very little addition, possibly quite standard, so it's done directly in the base addon.

@Tecnativa TT17678

### DIFF
--- a/l10n_eu_nace/views/res_partner.xml
+++ b/l10n_eu_nace/views/res_partner.xml
@@ -12,6 +12,7 @@
             <xpath expr="//field[@name='category_id']" position="after">
                 <field name="nace_id"
                        options="{'always_reload': True}"/>
+                <field name="secondary_nace_ids" widget="many2many_tags"/>
             </xpath>
         </field>
     </record>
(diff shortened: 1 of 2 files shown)


sbejaoui (approved):

l10n_eu_nace/views/res_partner.xml
```diff
@@ -12,6 +12,7 @@
             <xpath expr="//field[@name='category_id']" position="after">
                 <field name="nace_id"
                        options="{'always_reload': True}"/>
+                <field name="secondary_nace_ids"/>
```
pedrobaeza: Use `widget="many2many_tags"`

pedrobaeza (changes requested):

l10n_eu_nace/views/res_partner.xml
yajo: Changed.

pedrobaeza (approved):

PR #91, opened 2020-07-22
[12.0][ADD] allowance codes

badbole:
Base UNECE codes for allowances/discounts to be used in other modules


badbole:
@astirpe - i see you respond, so please, could you check this as prerequisite for other modules in edi repo realated to allowances in UBL standard... tnx

allowance_unece/models/unece_code_list.py
```diff
@@ -0,0 +1,13 @@
...
+# @author: Davor Bojkic <davor.bojkic@decod.io>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models, fields
+
+
+class UneceCodeList(models.Model):
+    _inherit = 'unece.code.list'
+
+    type = fields.Selection(selection_add=[
```
astirpe, line 11: Every time you make a `selection_add`, you should provide an `uninstall_hook` to provide clean uninstall. See for example https://github.com/odoo/odoo/issues/45240#issuecomment-585727789. You can find many examples of uninstall hook in OCA modules.

allowance_unece/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020 Decodio Applications (<http://decod.io>)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Base UBL allowance",
+    'version': "12.0.1.0.0",
+    "author": "Decodio Applications, Odoo Community Association (OCA)",
+    "development_status": "RC",
```
astirpe: Here are the allowed values for `development_status`:

https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L8

allowance_unece/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "summary": "Discount codes for UBL",
+    'website': "https://github.com/OCA/edi",
```
astirpe:
```suggestion
    'website': "https://github.com/OCA/community-data-files",
```

allowance_unece/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020 Decodio Applications (<http://decod.io>)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Base UBL allowance",
+    'version': "12.0.1.0.0",
+    "author": "Decodio Applications, Odoo Community Association (OCA)",
+    "development_status": "RC",
+    "summary": "Discount codes for UBL",
```
astirpe, line 7: Is this correct?

astirpe (changes requested):
I made only a code review (no functional review).

allowance_unece/__manifest__.py
badbole: as for discount/allowance codes .. yes.. 
defined by unece, comp'lete codes list here:
http://www.unece.org/trade/untdid/d97b/uncl/uncl5189.htm

will make other requested modifications and push soon

allowance_unece/models/unece_code_list.py
badbole: i agree that selection-add can be tricky, but none of modules extending unece_code_list model have such hook.. 
it adds a data sorted according to field selection.. if module is uninstalled, all functional domains on model unece_code_list used by module are not functional, so even if data is left after uninstall process it will not imact on anything... 
check in modules : account_tax_unece, account_payment_unece
still thinking we need uninstall hook?

astirpe: You are right. This selection is not functionally used, so it doesn't cause any error.

astirpe (approved):

PR #92, opened 2020-08-07
[FIX] l10n_eu_nace: Bad type in pre migration with boolean attribute

flotho:
FIX + IMP : more readable query for updating partners with new xml_id regarding old xml_id

### DIFF
--- a/l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
+++ b/l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
@@ -8,22 +8,22 @@ def migrate(cr, version):
         """
         WITH imd_new AS(
             SELECT 
-                imd_new.name AS xml_id, 
+                imd_new.name AS xml_id,
                 imd_new.res_id AS new_res_id
             FROM ir_model_data AS imd_new
-            WHERE imd_new.name NOT LIKE 'old_%' 
+            WHERE imd_new.name NOT LIKE 'old_%'
             AND imd_new.module = 'l10n_eu_nace'
         ),
         imd_old AS(
-            SELECT 
-                substring(imd_old.name, 5) AS xml_id, 
+            SELECT
+                substring(imd_old.name, 5) AS xml_id,
                 imd_old.res_id AS old_res_id
             FROM ir_model_data AS imd_old
-            WHERE imd_old.name LIKE 'old_%' 
+            WHERE imd_old.name LIKE 'old_%'
             AND imd_old.module = 'l10n_eu_nace'
         ),
         match_old_new AS(
-            SELECT 
+            SELECT
                 imd_new.xml_id,
                 imd_new.new_res_id,
                 imd_old.old_res_id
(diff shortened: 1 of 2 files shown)


flotho:
HI  @sbejaoui  and @Yajo  ,

I propose an easy fix here, thanks for your review.

regards

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -6,13 +6,44 @@
...
+            FROM res_partner AS rp
+            INNER JOIN res_partner_res_partner_category_rel AS rpc ON rpc.partner_id = rp.id
+            INNER JOIN match_old_new ON match_old_new.old_res_id = rpc.category_id
+            
+        )
+        UPDATE res_partner rp
+        SET nace_id = partner_with_old_nace.new_res_id
+        FROM partner_with_old_nace
+        WHERE rp.id = partner_with_old_nace.part_id;
             """
```
yajo: Is this code basically doing the same thing as before? If so, my vote goes for reverting this change. SQL is usually hard to interpret (mentally), but it's easier to interpret 5 lines than 35.

l10n_eu_nace/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -6,7 +6,7 @@ def migrate(cr, version):
     cr.execute(
         """
         UPDATE ir_model_data
-        SET name=concat('old_', name), noupdate=1
+        SET name=concat('old_', name), noupdate=TRUE
```
yajo, line 9: Is this really a bug? Do you have migration logs on the subject?

I mean, I'm :+1: with this change anyways, but I'm curious about this for future migrations.

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
flotho: Hi @Yajo 
Actually the previous query was failing because :
* it was addressing a field category_id on res_partner that doesn't exists.
* it doesn't taking into account the fact that category_id is a many2many fields

Maybe I'm totally wrong on the analysis of the query but as it was failing on my pgsql I debugged this and propose something "more clear", in my opinion,

regards

l10n_eu_nace/migrations/12.0.2.0.0/pre-migration.py
flotho: PgSQL 12 was insulting me on this that's why I proposed this patch.
Actually we're executing directly a SQL query and either or odoo is throwing : 
```
ERROR: ERREUR:  la colonne « noupdate » est de type boolean mais l'expression est de type integer
LINE 3:         SET name=concat('ol2_', name), noupdate=1
                                                       ^
HINT:  Vous devez réécrire l'expression ou lui appliquer une transformation de type.
SQL state: 42804
Character: 79
```

l10n_eu_nace/migrations/12.0.2.0.0/post-migration.py
yajo: OK, this is a bit too much verbose IMHO, but after reading all closely, it's fixing the bug, so :+1:

yajo (approved):

yajo:
It seems you have linter failures: https://travis-ci.org/github/OCA/community-data-files/jobs/715765545#L321-L331

PR #94, opened 2020-09-16
[IMP] - multiple codes for a bank

sbejaoui:
This PR make the bank code a one2many field to cover the case where the same bank have different national bank code in the same country.


sbejaoui:
cc/ @gva-acsone

PR #81, opened 2020-01-31
[13.0] [MIG] base_iso3166

ThomasBinsfeld:

PR #95, opened 2020-10-01
[13.0][MIG] base_iso3166: Migration to 13.0

ernestotejeda:
Supersedes https://github.com/OCA/community-data-files/pull/81

cc @Tecnativa TT21617


victoralmau (approved):
Maybe need to add more test to pass codecov.

base_iso3166/models/res_country.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2016 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# Copyright 2017  Creu Blanca <www.creublanca.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+try:
+    import pycountry
+except ImportError:
+    pass
```
joao-p-marques: Why this here? If you added `pycountry` to the dependencies and to `requirements.txt` this should always work, right? Besides, will the module work without `pycountry` installed? If not, maybe it should give an error anyway...

joao-p-marques (commented):
Just some small comments. Everything else LGTM :+1:

base_iso3166/models/res_country.py
ernestotejeda: This comes from previous versions but I think that `try: .... except: ...` could be removed since there is no specific behavior defined for when the library is not found. Do you agree @pedrobaeza

pedrobaeza: Yes, now it's safe to not protect imports, as Odoo handles this correctly.

joao-p-marques: :+1:

ernestotejeda:
> Maybe need to add more test to pass codecov.

@victoralmau , if you have any idea how to cover the only line that is not covered, help me with that, because I don't know how to do it. :(

pedrobaeza:
The problem here is that the `try` lines are for handling specific `pycountry` library versions that change the API constantly, so indeed it's very difficult to be covered by tests.

pedrobaeza:
Please give credits here in the contributors file to @ThomasBinsfeld as he started the migration.

joao-p-marques (approved):

ernestotejeda:
> Please give credits here in the contributors file to @ThomasBinsfeld as he started the migration.

@pedrobaeza , changes done.

FernandoRomera (approved):

PR #98, opened 2020-10-16
[14.0][MIG] base_unece

alexis-via:

PR #99, opened 2020-10-16
[14.0][MIG] account_tax_unece

alexis-via:


alexis-via:
Requires #98

account_tax_unece/__manifest__.py
```diff
@@ -1,22 +1,23 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 # @author Alexis de Lattre <alexis.delattre@akretion.com>
 
 {
     "name": "Account Tax UNECE",
-    "version": "13.0.1.0.0",
+    "version": "14.0.1.0.0",
     "category": "Accounting & Finance",
     "license": "AGPL-3",
+    "development_status": "Production/Stable",
```
pedrobaeza, line 10: I would say this one can be already `Mature`

account_tax_unece/data/unece_tax_categ.xml
```diff
@@ -20,57 +20,62 @@ Source : https://www.unece.org/fileadmin/DAM/trade/untdid/d16b/tred/tred5305.htm
...
     </record>
     <record id="tax_categ_c" model="unece.code.list">
         <field name="type">tax_categ</field>
         <field name="code">C</field>
         <field name="name">Duty paid by supplier</field>
-        <field
-            name="description"
-        >Duty associated with shipment of goods is paid by the supplier; customer receives goods with duty paid.</field>
+        <field name="description">
+            Duty associated with shipment of goods is paid by the supplier; customer
```
pedrobaeza: Are all of these changes done by pre-commit? If so, they must be in the pre-commit commit, not here.

pedrobaeza (changes requested):
It seems there are changes in the migration commit that belongs to pre-commit. Also, the pre-commit commit message is not the one stated in the migration guide.

account_tax_unece/data/unece_tax_categ.xml
alexis-via: What is a pre-commit commit ? Does it mean that you would have prefered to see in a separate commit ?

pedrobaeza: Please check https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0#technical-method-to-migrate-a-module-from-130-to-140-branch where migration steps including pre-commit are included

astirpe: It is possible to skip the automatic reformatting of a particular file by adding its path in the `.prettierignore`, like we did here: https://github.com/OCA/l10n-netherlands/blob/13.0/.prettierignore

pedrobaeza: The reformatting is good and should be done here, but on its separate commit according the procedure.

astirpe: The reformatting is not good.

Here for example we have:

```xml
        <field name="description">
            Duty associated with shipment of goods is paid by the supplier
```

which however does not behave the same as:


```xml
        <field
           name="description"
          >Duty associated with shipment of goods is paid by the supplier
```

If you want to realize it, open the unece.code.list tree view: you will notice that the text in the description column is misaligned in most of the records.

Since the automatic reformatting is forcing such unwanted changes, just include the path of the files in the .prettierignore so that the reformatting will be not executed.

astirpe (changes requested):
See my comment https://github.com/OCA/community-data-files/pull/99#discussion_r510832164

account_tax_unece/data/unece_tax_categ.xml
pedrobaeza: OK, then it's good to know and indeed it affects in this case.

@Yajo is this something related with ignore-whitespace things?

yajo: Fix in https://github.com/OCA/oca-addons-repo-template/pull/35.

rvalyi:
@alexis-via so now that the pre-commit hook was fixed with a less aggressive prettier XML plugin, (https://github.com/OCA/oca-addons-repo-template/issues/34) I think you should redo (and amend) your PR and it should fix the wrongly reformatted XML code mentioned above.

As for the pre-commit thing, aside from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0#technical-method-to-migrate-a-module-from-130-to-140-branch you can also read an introduction here https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97
It's easy to set up on your local PC with pre-commit install / pre-commit uninstall and when last glitches will be sorted out at the OCA level it may really improve the average OCA code readability.

astirpe (approved):

yajo:
I think what you should do is this instead.

```bash
git reset --hard 88af9af61af15817476cd84a67198f9abc2a946b

# Assumming `upstream` is the remote that points to OCA's repo
git pull upstream 14.0 --rebase

pre-commit run -a
git commit -am '[IMP] account_tax_unece: run pre-commit'

git cherry-pick 34c423e2f9fdd2f149dad17805cd8ca36b72b728

# Now fix conflicts, if there are any (there'll be). After that's done:
git cherry-pick --continue
git push --force
```

I'm sorry this problem hit you

phuctranfxvn:
Hello, I see that you exclude the [test_account_invoice](https://github.com/OCA/community-data-files/blob/13.0/account_tax_unece/tests/test_account_invoice.py) from your test case. Did you do it accidentally or for a specific reason ?
I think you should include the setup/account_tax_unece in your PR also.

StefanRijnhart:
@alexis-via will you follow-up? Or if you'd like, I can post a superseding PR.

alexis-via:
@StefanRijnhart Don't hesitate to supersede the PR :)

StefanRijnhart:
Thanks! Continued here: https://github.com/OCA/community-data-files/pull/103

PR #106, opened 2021-02-25
12.0 mig product uom unece

Kev-Roche:


StefanRijnhart:
Module seems to be already ported, albeit renamed to `uom_unece`: https://github.com/OCA/community-data-files/tree/12.0/uom_unece

Kev-Roche:
@StefanRijnhart 
You're right, sorry about it !

PR #109, opened 2021-04-27
[14.0][MIG] l10n_eu_product_adr

StefanRijnhart:
```
14.0.1.0.0 (2021-05-04)
~~~~~~~~~~~~~~~~~~~~~~~
* [BREAKING] Gather dangerous goods attributes on a new model adr.goods,
  and replace all product attributes with a Many2one on this model.
  Remove all other unused or unclear attributes.
* [ADD] XML data for the dangerous goods table from the 2019 edition,
  including translations in 3 languages.
```


StefanRijnhart:
@i-vyshnevska I have observed some points to improve this module; https://github.com/OCA/community-data-files/pull/109/files#diff-d967fda5116ea9b63e06dd49e2557fbfc802b8b6476598adcc50acd9d8b890b5R49-R73. Do you agree, or are there good reasons for some of the current design choices? What part of the specifications does `great.class` refer to?

i-vyshnevska:
@StefanRijnhart your comments looks reasonable, what I can say would be good to have one more POV as work was based on just one customer requirements, also related report appeared to be more specific for the country

simahawk:
@StefanRijnhart is this ready now?

StefanRijnhart:
@simahawk thank you for your interest. I've now pushed the latest version as accepted by our customer. I believe it is more correct as it allows for multiple entries in the dangerous goods table with the same UN number, distinguished by different packing instructions or limited quantities. It does mean however that the already very bare migration from the previous version was invalidated entirely so I removed it.

l10n_eu_product_adr/models/adr_class.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2021 Stefan Rijnhart <stefan@opener.amsterdam>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+from odoo.osv.expression import AND
+
+
+class AdrClass(models.Model):
+    _name = "adr.class"
```
i-vyshnevska, line 8: you can add a migration here as we renamed clasess

l10n_eu_product_adr/models/adr_label.py
```diff
@@ -0,0 +1,29 @@
...
+
+    name = fields.Char(required=True, translate=True)
+    code = fields.Char(required=True)
+    image = fields.Binary(required=True)
+    class_id = fields.Many2one(comodel_name="adr.class", string="Class", required=True)
+
+    def unlink(self):
+        """Restrict removal of labels in use"""
+        for label in self:
+            if label.class_ids:
```
i-vyshnevska:
```suggestion
            if label.class_ids:
```
I see class_id not class_ids and `ondelete='restrict'` should solve this issue

i-vyshnevska (commented):
in new version we not distinguish good and waste

l10n_eu_product_adr/models/adr_class.py
StefanRijnhart: Adding *only* the migration of this class would not be wrong, but not very useful by itself I'm afraid.

l10n_eu_product_adr/models/adr_label.py
StefanRijnhart: Thanks, I missed this at some point during the refactoring. I've changed this to check the linked goods, and added tests for this method and others.

StefanRijnhart:
> in new version we not distinguish good and waste

@i-vyshnevska sorry, I don't know what you mean by this comment. Can you clarify?

l10n_eu_product_adr/scripts/import_adr_multilang_xlsx.py
```diff
@@ -0,0 +1,450 @@
...
+# Transforms the spreadsheet found at https://www.cepa.be/wp-content/uploads\
+#    /ADR_2019_BijlageA_deel3_Tabel_A_EXCEL_FORMAAT.xlsx
+# to adr.goods XML data.
+
+import re
+import sys
+from collections import defaultdict
+
+from lxml import etree
+from openpyxl import load_workbook  # pylint: disable=W7936
```
mmequignon, line 12: Please wrap this import with a `try`, `except`
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#32imports

l10n_eu_product_adr/models/adr_goods.py
```diff
@@ -0,0 +1,102 @@
...
+    def _check_un_number(self):
+        for goods in self:
+            if len(goods.un_number) != 4:
+                raise ValidationError(
+                    _(
+                        "UN Number %s is invalid because it does not have "
+                        "a length of 4."
+                    )
+                    % goods.un_number
+                )
```
mmequignon, line 78: We had to implement that as well. I don't recall why, but the our customer has some cases where the unnumber isn't a 4 digits long string. Here's where I implemented the solution we agreed on https://github.com/OCA/delivery-carrier/compare/0278821af40dc8fef2fc2518f53038a37bb07deb..0758137ff35a9a0e3170b641d566749de802df00

l10n_eu_product_adr/models/product_template.py
```diff
@@ -1,118 +1,43 @@
...
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
-    is_dangerous = fields.Boolean(string="Dangerous product")
-    dangerous_class_id = fields.Many2one(
-        "product.dangerous.class", string="Dangerous class"
+    is_dangerous = fields.Boolean(help="This product belongs to a dangerous class")
+    adr_goods_id = fields.Many2one("adr.goods", "Dangerous Goods")
+    adr_class_id = fields.Many2one(
+        "adr.class", related="adr_goods_id.class_id", readonly=True
```
mmequignon: Since the dangerous attributes can vary from a variant to another, we moved those fields from `product.template` to `product.variant`.
see this PR https://github.com/OCA/community-data-files/pull/112

Could you please do the same?

mmequignon (approved):
functional tests ok, even if it'll break a lot of module which depends on this one ^^'
However, @i-vyshnevska knows better about this.

i-vyshnevska:
> > in new version we not distinguish good and waste
> 
> @i-vyshnevska sorry, I don't know what you mean by this comment. Can you clarify?

previously we had two classes to distinct dangerous "good" and "waste" (products which producing dangerous staff after they used) they treated differently in reports etc

Highcooley:
> > > in new version we not distinguish good and waste
> > 
> > 
> > @i-vyshnevska sorry, I don't know what you mean by this comment. Can you clarify?
> 
> previously we had two classes to distinct dangerous "good" and "waste" (products which producing dangerous staff after they used) they treated differently in reports etc

since our customer does not work with dangerous waste at the moment, I think we can ignore this.

l10n_eu_product_adr/models/adr_goods.py
```diff
@@ -0,0 +1,102 @@
...
+    label_ids = fields.Many2many(
+        comodel_name="adr.label",
+        string="Labels",
+        help=(
+            "The labels/placards (see 5.2.2.2 and 5.3.1.7) that have to be affixed "
+            "to packages, containers, tank-containers, portable tanks, MEGCs and "
+            "vehicles.)"
+        ),
+    )
+    limited_quantity = fields.Char()
```
mmequignon: This one should be different
```suggestion
    limited_quantity = fields.Float()
    limited_quantity_uom = fields.Many2one("uom.uom")
```

l10n_eu_product_adr/models/product_template.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    adr_classification_code = fields.Char(
+        related="adr_goods_id.classification_code", readonly=True
+    )
+    adr_label_ids = fields.Many2many(
+        "adr.label", related="adr_goods_id.label_ids", readonly=True
+    )
```
mmequignon, line 32:
```suggestion
    )
    limited_quantity = fields.Boolean("Limited Qty", compute="_compute_limited_quantity", store=True)
```
With the compute method like so
```python
@api.onchange("adr_goods_id", "adr_goods_id.limited_quantity_uom", "adr_goods_id.limited_quantity", "uom_id", )
def _compute_limited_quantity(self):
    for record in self:
        compare product's uom_qty with adr.goods uom_qty:
        product.limited_qty = adr.good uom_qty < product uom_qty
```

We should also ensure than the `adr.label` uom has the same uom family as the product one.

mmequignon (commented):
@StefanRijnhart 
We discussed internally about this, and it does not provide a way to tell that a product is valid or not, regarding the limited_qty.
I provided you a possible solution in the comments.
Thanks!

l10n_eu_product_adr/scripts/import_adr_multilang_xlsx.py
StefanRijnhart: Ah, but this is not necessary in recent versions of Odoo. The linked page (now) mentions this as well:
```
This rule doesn't apply neither to Odoo >= v12, as an unmet dependency in an uninstalled module doesn't block the service thanks to this commit:
```

l10n_eu_product_adr/models/adr_goods.py
StefanRijnhart: It would be interesting to hear when such a number can appear, because no such number appears in any of the specs I've seen!

l10n_eu_product_adr/models/product_template.py
StefanRijnhart: Alright, I will take care of that.

l10n_eu_product_adr/models/adr_goods.py
StefanRijnhart: OK. I will not commit your suggestion directly because of the implications, but I will have a look at this.

l10n_eu_product_adr/models/product_template.py
StefanRijnhart: I'm sorry, but my customer is a bulk trader. They trade in pallets, or gallons, and the individual subpackaging is encoded in the product name rather than the UoM of the product. So what suits your requirements is not generally valid.

mmequignon:
@StefanRijnhart 
Hello Stefan, I see that you put this PR back in progress.
Could you please list what you will do, and give us a raw estimate of how many time it will take?
Thank you!

StefanRijnhart:
@mmequignon I tried to indicate in my comments what I was going to refactor:

* https://github.com/OCA/community-data-files/pull/109#discussion_r745969757
* https://github.com/OCA/community-data-files/pull/109#discussion_r745970606

I hope to finish it this week on Monday or Tuesday.

l10n_eu_product_adr/models/adr_goods.py
StefanRijnhart: See new commit "[IMP] l10n_eu_product_adr: distinct uom for limited quantity" in this PR for this.

l10n_eu_product_adr/models/product_template.py
StefanRijnhart: See new commit `[RFR] l10n_eu_product_adr: allow different adr goods per variant`

mmequignon (approved):

mmequignon:
We checked this with the customer, and there's a few missing fields:
- hazard_ind https://github.com/OCA/community-data-files/blob/13.0/l10n_eu_product_adr/models/product_product.py#L50
- packaging_group https://github.com/OCA/community-data-files/blob/13.0/l10n_eu_product_adr/models/product_product.py#L52-L54
- envir_hazardous https://github.com/OCA/community-data-files/blob/13.0/l10n_eu_product_adr/models/product_product.py#L31-L33 (for this one, I'll ask how to determine its value)

[adr-tool](https://adr-tool.com/1339/un-1993)

simahawk:
@i-vyshnevska can you update your review?
@StefanRijnhart do you think we can merge this?

StefanRijnhart:
@simahawk that would be good. The Camptocamp team can then propose follow-up PRs for improvements as they see fit.

l10n_eu_product_adr/__manifest__.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2019 Camptocamp SA
+# Copyright 2021 Opener B.V.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "ADR Dangerous Goods",
+    "summary": "Allows to set appropriate danger class and components",
+    "version": "14.0.1.1.0",
```
i-vyshnevska, line 7:
```suggestion
    "version": "14.0.1.0.0",
```

l10n_eu_product_adr/models/adr_goods.py
```diff
@@ -0,0 +1,106 @@
...
+                raise ValidationError(
+                    _(
+                        "UN Number %s is invalid because it does not have "
+                        "a length of 4."
+                    )
+                    % goods.un_number
+                )
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
```
i-vyshnevska, line 81: do you need to override name_search if un_number is already in name_get?

l10n_eu_product_adr/models/product_template.py
```diff
@@ -0,0 +1,102 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    adr_goods_on_variants = fields.Boolean(
+        compute="_compute_adr_goods_on_variants",
+        help="Indicates whether the adr configuration is different for each variant.",
+    )
+    is_dangerous = fields.Boolean(
+        related="product_variant_ids.is_dangerous",
```
i-vyshnevska, line 16: I would rather forbid having under one template dangerous and not dangerous product, it is hard to imagine such situation

l10n_eu_product_adr/models/product_template.py
```diff
@@ -0,0 +1,102 @@
...
+        help="This product belongs to a dangerous class",
+    )
+    adr_goods_id = fields.Many2one(
+        "adr.goods",
+        "Dangerous Goods",
+        related="product_variant_ids.adr_goods_id",
+        readonly=False,
+    )
+    adr_class_id = fields.Many2one(
+        "adr.class", related="product_variant_ids.adr_goods_id.class_id"
```
i-vyshnevska, line 27: same hare if products have different classes what class will have a template?

l10n_eu_product_adr/models/stock_picking.py
```diff
@@ -0,0 +1,19 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    adr_points = fields.Float(
+        compute="_compute_adr_points",
+        string="ADR Points",
+        digits="Product Unit of Measure",
+    )
+
+    @api.depends("move_lines")
```
i-vyshnevska:
```suggestion
    @api.depends("move_lines.product_id", "move_lines.product_uom_qty)
```

l10n_eu_product_adr/scripts/import_adr_multilang_xlsx.py
```diff
@@ -0,0 +1,485 @@
...
+    code = parse_classification_code(row)
+    etree.SubElement(
+        record, "field", attrib={"name": "classification_code"}
+    ).text = code
+
+
+def parse_transport_category(row):
+    value = row[index["transport_category"]]
+    if value is None:
+        if "VERVOER VERBODEN" in str(row):
```
i-vyshnevska, line 309: this is very customer specific

l10n_eu_product_adr/__manifest__.py
StefanRijnhart: I'm sorry, but I changed the datamodel on your request and added the related migration for my customer and for anyone else who might have been running this code so I'm not taking this suggestion.

l10n_eu_product_adr/models/product_template.py
StefanRijnhart: That seemed to me to be a necessary consequence of @mmequignon's request to allow different settings per product variant. If it is not allowed, it is not possible to extend the number of variants by adding product attribute options on the template because the actual dangerous good is not available to the Odoo mechanism that expands the number of variants.

l10n_eu_product_adr/models/product_template.py
StefanRijnhart: It will be assigned the class of the first variant. This is necessary to keep compatibility for my customer who does not have different classes per variant.

l10n_eu_product_adr/scripts/import_adr_multilang_xlsx.py
StefanRijnhart: This is not customer specific, it is specific to the sheet that contains the ADR data. It is made available by a Belgian organization, and as a consequence it contains a limited amount of Dutch terminology.

l10n_eu_product_adr/models/adr_goods.py
StefanRijnhart: The un_number is included in the representation of the record returned by `name_get`, but it is not included in the record's name so I believe the answer is yes.

l10n_eu_product_adr/__manifest__.py
sebalix: Let's keep this version then.

sebalix:
I think we cannot merge this as-is because:

- some fields are still missing, see the following comment: https://github.com/OCA/community-data-files/pull/109#issuecomment-975698882 (if we do this in a new PR that would mean that users migrating from 13.0 to 14.0 with this version will get the missing columns dropped from the DB until we add the missing fields in a new version, is it ok to do this?)
- we should rebase & merge https://github.com/StefanRijnhart/community-data-files/pull/1 to improve the migration script

StefanRijnhart:
I can add hazard_ind and packaging_group, as these are also present in the sheet that I import the classes from. I was waiting if more information about environmentally hazardous was coming, but I think that maybe it is obsolete
image
(screenshot from https://adrbook.com/en/2017/ADR/2.2.9.1.9)

l10n_eu_product_adr/security/ir.model.access.csv
```diff
@@ -0,0 +1,9 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_adr_goods_user,access_adr_goods,model_adr_goods,stock.group_stock_user,1,0,0,0
+access_adr_goods_manager,access_adr_goods,model_adr_goods,base.group_erp_manager,1,1,1,1
+access_adr_class_user,access_adr_class,model_adr_class,stock.group_stock_user,1,0,0,0
+access_adr_class_manager,access_adr_class,model_adr_class,base.group_erp_manager,1,1,1,1
+access_adr_label_user,access_adr_label,model_adr_label,stock.group_stock_user,1,0,0,0
+access_adr_label_manager,access_adr_label,model_adr_label,base.group_erp_manager,1,1,1,1
+access_adr_packing_instruction_user,access_adr_packing_instruction,model_adr_packing_instruction,stock.group_stock_user,1,0,0,0
```
TDu: Could the read access be set on the `base.group_user` instead ?
I see access error for non inventory user opening product pages.
Same question for all read access in this file.

StefanRijnhart: Thanks for your comment, I'll check how to prevent the access error.

sebalix: Indeed, we are facing some access errors when encoding an SO because of this new data model `adr.label`. @StefanRijnhart did you check it or you prefer we open a PR on your branch?

StefanRijnhart: @sebalix I have a fix in mind for the product form, but if you run into problems creating an SO then it would be interesting to have a traceback.

StefanRijnhart: Fix pushed. Rather than extending the access rights, I've added groups on the product views to prevent the accdess error. I also apply `compute_sudo` to the computation of stock move's adr points, I hope this helps when validating a sales order.

rousseldenis:
@StefanRijnhart What's the status of this ?

bosd:
Can the generation of a new Runboat be triggered?

marylla:
> Can the generation of a new Runboat be triggered?

@rousseldenis @StefanRijnhart Do you know how to do it? On details it just says {"detail":"Not Found"}...

rousseldenis:
@StefanRijnhart Could your rebase ?

StefanRijnhart:
I've rebased. As for the status, I'm considering to let this one slip because as expected this does not align well with the team that provided the previous implementation, and I think the communication has not been easy for either side. Also, I'm not working for the project that required this anymore and I'm a bit of a hippie so I prefer to work with bicycles rather than with dangerous goods anyway :grin:

marylla:
@bosd The runboat is now working, so you can test it. :) I would too but I don't know enough about this topic to test it properly.

bosd (approved):
LGTM!

marylla:
@jgrandguillaume Can we please merge this one?

bosd:
Maybe @OCA/community-data-files-maintainers can help

simahawk:
FYI On our side we are waiting for @mmequignon to get back to this for a final review.
We have many other modules depending (and blocked) on this and we need to check how they are impacted after the last changes.
To name some:
* https://github.com/OCA/wms/pull/334
* https://github.com/OCA/delivery-carrier/pull/441
* https://github.com/OCA/stock-logistics-workflow/pull/937

mmequignon (approved):

PR #111, opened 2021-06-03
[IMP] base_iso3166: Only add numeric if it exists

etobella:
There are historic countries without numeric

@OCA/tools-maintainers


albariera (approved):

LoisRForgeFlow (approved):
:+1:

PR #113, opened 2021-08-20
[13.0][IMP] dangerous product add full name

TDu:
For the product.dangerous.class Having the name and the code
in the UI when looking at a product is helpfull to the user.

### DIFF
--- a/l10n_eu_product_adr/models/product_dangerous.py
+++ b/l10n_eu_product_adr/models/product_dangerous.py
@@ -6,10 +6,10 @@ from odoo import api, fields, models
 class ProductDangerousClass(models.Model):
     _name = "product.dangerous.class"
     _description = "Product Dangerous Class"
-    _rec_name = "name_full"
+    _rec_name = "complete_name"
 
     name = fields.Char(required=True, translate=True)
-    name_full = fields.Char(compute="_compute_name_full")
+    complete_name = fields.Char(compute="_compute_complete_name")
     code = fields.Char(required=True)
     image = fields.Binary(string="Icon", required=True)
     class_type_id = fields.Many2one(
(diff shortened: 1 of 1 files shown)


l10n_eu_product_adr/models/product_dangerous.py
```diff
@@ -1,13 +1,15 @@
...
+from odoo import api, fields, models
 
 
 class ProductDangerousClass(models.Model):
     _name = "product.dangerous.class"
     _description = "Product Dangerous Class"
+    _rec_name = "name_full"
 
     name = fields.Char(required=True, translate=True)
+    name_full = fields.Char(compute="_compute_name_full")
```
sebalix: I would named it `complete_name` like in:

- [`stock.location`](https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_location.py#L30)
- [`product.category`](https://github.com/odoo/odoo/blob/aff7694c56185212c17ceaf358ad99b3d14dee8b/addons/product/models/product.py#L27)
- [`analytic.account.group`](https://github.com/odoo/odoo/blob/aff7694c56185212c17ceaf358ad99b3d14dee8b/addons/analytic/models/analytic_account.py#L46)
- `ir.ui.menu`
- `ir.model.fields`

`name_full` is not used at all in Odoo standard.

TDu: Yes and changed.

sebalix (approved):

l10n_eu_product_adr/models/product_dangerous.py
```diff
@@ -16,6 +18,11 @@ class ProductDangerousClass(models.Model):
         string="Dangerous Type",
     )
 
+    @api.depends("name", "code")
+    def _compute_complete_name(self):
+        for record in self:
+            record.complete_name = " ".join([record.code, record.name])
```
flotho, line 24: What about adding the class_type_id representation in the complete_name ?

flotho (approved):
LGTM

small comment

l10n_eu_product_adr/models/product_dangerous.py
TDu: Looking at the existing data, it seems to me that there would be a lot of duplicate information by adding the `class_type_id` name and not much gain.
The code is being added because from our experience people working with this data can quickly gather more information with the number in the code.

flotho (approved):

PR #115, opened 2021-10-04
[FIX] account_tax_unece: fix migration path from 13.0

sebalix:
Since `unece.code.list` records of type `date` have been removed on
14.0, the upgrade to 14.0 is causing a constraint issue when Odoo is
removing these XML data records from the database while they could still
be used by `account.tax` records through the `unece_due_date_id` field.


pedrobaeza:
The usual technique in OpenUpgrade is to be not destructive (as any other module may restore this date behavior and reuse the information), so the script should simply lift the constraint over such field. It's true that the records are removed, but you still have the correspondence applied on target records. That can be done with the openupgradelib method https://github.com/OCA/openupgradelib/blob/6d97dfc0af2f528889f4a491c71129ef0f7cf82f/openupgradelib/openupgrade.py#L2094

PR #135, opened 2022-07-01
Update of the translations of ``base_unece``, ``account_tax_unece`` and ``account_payment_unece``.

eferriere:
This is an update of the translations of ``base_unece``, ``account_tax_unece`` and ``account_payment_unece``.


pedrobaeza:
This is not correct picked and it's more than translations. Besides, translations are done on https://translation.odoo-community.org/

PR #150, opened 2022-12-22
[15.0][FIX] product_fao_fishing: Rename migration folder version

sergio-teruel:
cc @Tecnativa 

ping @stefan-tecnativa


pedrobaeza (approved):
Let's fast-track it

PR #151, opened 2022-12-23
[14.0][FIX] base_bank_from_iban: Format and get bank from the wizard

pedrobaeza:
New company bank accounts are not created from a special view of the journal, but on a specialized wizard.

The module hasn't been adapted to this situation, so the bank journals don't get the bank account formatted nor the bank to make SEPA payments.

We fix this including the code in the wizard, and removing the journal one that is no longer needed.

@Tecnativa


carlosdauden (approved):

pedrobaeza:
Let's fast-track the patch to get it ASAP:

/ocabot merge patch

PR #170, opened 2023-07-27
[IMP] base_bank_from_iban: automatically create banks with schwifty

len-foss:
Using Schwifty, we can automatically get bank information from a valid IBAN, i.e. BIC, Swift, code, and bank name.
This quick POC automatically create banks at creation/write of a bank account.

In practice the current code might be annoying if used on a DB where banks have already been created with incomplete information. I don't have have any relevant experience with that so I'd be interested to know the best approach to that issue. If used on a new DB, it should just simplify neatly the encoding process.

This could also go into an overriding module.

https://schwifty.readthedocs.io/en/latest

### DIFF
--- a/base_bank_from_iban/__manifest__.py
+++ b/base_bank_from_iban/__manifest__.py
@@ -12,5 +12,6 @@
     "depends": ["base_iban"],
     "development_status": "Mature",
     "data": ["views/res_bank_view.xml"],
+    "external_dependencies": {"python": ["schwifty"]},
     "installable": True,
 }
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Good idea!

carlosdauden (approved):

Abranes (approved):

ValentinVinagre:
@pedrobaeza @carlosdauden @len-foss @Abranes Hey,
I think that PR is merged with a wrong message in the commit. Could you change it?

pedrobaeza:
Indeed, I wasn't aware, but I'm afraid now that this can't be changed.

ValentinVinagre:
> Indeed, I wasn't aware, but I'm afraid now that this can't be changed.

the branch is protected and force cannot be used?

pedrobaeza:
Doing that will screw all the people that have already pulled this branch, forcing them to do a reset, so that's too much to do, and more having passed several weeks since the merge.

ValentinVinagre:
And a rebert and then commit again with the correct message?

pedrobaeza:
A more convoluted history... Let's fix this in v17. At the end, the commit is linked to this PR, where there is enough information.

PR #172, opened 2023-10-05
[16.0] base_unece + account_tax_unece: AGPL -> LGPL

alexis-via:
If all the contributors on these 2 modules agree, I propose to switch base_unece and account_tax_unece from AGPL to LGPL. This would allow some forks of localization modules, which are LGPL modules, to depend on account_tax_unece to have the fields for the unece codes on tax templates. Example of such a module: l10n_fr_oca https://github.com/OCA/l10n-france/pull/423

I am the initial and main author of those 2 modules. These modules have received contributions from @astirpe @pedrobaeza and @levkar ... are you guys ok with this licence change ?

### DIFF
--- a/account_tax_unece/__init__.py
+++ b/account_tax_unece/__init__.py
@@ -1,3 +1,1 @@
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
 from . import models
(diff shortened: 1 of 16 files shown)


pedrobaeza:
OK, no problem on my side.

astirpe (approved):
OK for me!

alexis-via:
Thanks for your answers ! @levkar are you OK with this change ?

lk-eska:
Yes sure, we want to use it that way too. Thanks for asking.

alexis-via:
@lk-eska thanks for your support on this change

PR #176, opened 2023-10-29
[16.0][MIG] l10n_eu_nace

Abranes:
Standard migration to 16.0


flotho:
hum... runboat not available! issued a 404

Abranes:
> hum... runboat not available! issued a 404

It works for me, can you try it one more time?

flotho (approved):
functionnal test OK,
LGTM

iziwas:
Hi,

Thanks for the PR request. 
Functional tests are OK for me too. However, I have a warning about parent_path field in res.parter.nace : "parent_path field on model 'res.partner.nace' should have unaccent disabled. Add `unaccent=False` to the field definition."

Is it possible to fix it before please ?

yajo (changes requested):
After analyzing this module, I think it needs a refactor. Please let me explain.

This module introduces a duplicate structure in Odoo. [Odoo CE already has a `res.partner.industry` model](https://github.com/odoo/odoo/blob/13c595b342fbf09e188c5fab99724f4265a94317/odoo/addons/base/models/res_partner.py#L1112-L1119). In OCA, [the `partner_industry_secondary` module](https://github.com/OCA/partner-contact/tree/16.0/partner_industry_secondary) implements:
- Parent-child relationship.
- Many2many relation between `res.partner` and `res.partner.industry`.

So, if this module depends on `partner_industry_secondary`, which is already migrated to v16, we can drop most of the code. This would just contain the NACE codes in CSV, which would be declared in the manifest and loaded at install; period.

IMHO the migration is probably the best opportunity to do it, as this involves structural changes in the data model. It would mean directly dropping all xml and python code from here, so I think it's a simple refactor that'd be very helpful.

FWIW the only "complex" thing would be a migration script, which wouldn't be very complex anyways.

WDYT?

Rad0van:
> After analyzing this module, I think it needs a refactor. Please let me explain.
> 
> This module introduces a duplicate structure in Odoo. [Odoo CE already has a `res.partner.industry` model](https://github.com/odoo/odoo/blob/13c595b342fbf09e188c5fab99724f4265a94317/odoo/addons/base/models/res_partner.py#L1112-L1119). In OCA, [the `partner_industry_secondary` module](https://github.com/OCA/partner-contact/tree/16.0/partner_industry_secondary) implements:
> 
> * Parent-child relationship.
> * Many2many relation between `res.partner` and `res.partner.industry`.
> 
> So, if this module depends on `partner_industry_secondary`, which is already migrated to v16, we can drop most of the code. This would just contain the NACE codes in CSV, which would be declared in the manifest and loaded at install; period.
> 
> IMHO the migration is probably the best opportunity to do it, as this involves structural changes in the data model. It would mean directly dropping all xml and python code from here, so I think it's a simple refactor that'd be very helpful.
> 
> FWIW the only "complex" thing would be a migration script, which wouldn't be very complex anyways.
> 
> WDYT?

Looking at the modules I totally agree with you.

pedrobaeza:
The only problem I see is that the industry model already contains other records that are not NACE ones.

Rad0van:
> The only problem I see is that the industry model already contains other records that are not NACE ones.

Which ones? By quickly looking (at least for v16) it seems right to me...

pedrobaeza:
I mean these ones:

https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/data/res_partner_data.xml#L57-L160

Rad0van:
> I mean these ones:
> 
> https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/data/res_partner_data.xml#L57-L160

Yes, but these are the main existing NACE groups. See https://en.wikipedia.org/wiki/Statistical_Classification_of_Economic_Activities_in_the_European_Community or https://nacev2.com/en

pedrobaeza:
Ok then. Just a proper mapping should be done for recognizing them as the parents.

yajo:
`partner_industry_secondary` makes the parent-child relationship possible, so it shouldn't be a big problem, apart from a relatively simple migration script.

cvinh:
Hello @Abranes, are you going to change your PR according to comments above ?

cvinh:
I think that secondary industry is meant to allow several industries to the same partner (see that we cannot add a secondary industry that is already in the main industry : https://github.com/OCA/partner-contact/blob/23c2d101e08c36f3909091c5f488110ce45e14cc/partner_industry_secondary/models/res_partner.py#L31)
BTW res.partner.industry has a parent_id field
To my opinion, I would not inherit partner_industry_secondary but add the NACE codes to res_partner_industry and choose the right parent id from Odoo's datas

edlopen:
I have an [open PR](https://github.com/OCA/community-data-files/pull/196) that uses the res.parter.industry as @yajo said [here](https://github.com/OCA/community-data-files/pull/176#pullrequestreview-1749999335). And also updated the module so that it uses the new ShowVoc Endpoint as the RAMON servers are deprecated.

yajo:
Thanks! Closing here, as the other PR is more suited for merge.

rafaelbn:
Migrated in https://github.com/OCA/community-data-files/pull/196

PR #179, opened 2023-11-18
[17.0][MIG] base_iso3166: Migration to 17.0

ramiadavid:


pedrobaeza:
/ocabot migration base_iso3166

Last commit name is not correct.

ramiadavid:
@pedrobaeza Thanks, it's already fixed

base_iso3166/static/description/index.html
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: Why removing this?

base_iso3166/views/country_view.xml
```diff
@@ -16,7 +15,7 @@
         <field name="name">Country tree (with ISO 3166-1 alpha-3)</field>
         <field name="model">res.country</field>
         <field name="inherit_id" ref="base.view_country_tree" />
-        <field name="type">form</field>
+        <field name="type">tree</field>
```
pedrobaeza: Isn't this line obsolete?

base_iso3166/static/description/index.html
ramiadavid: `DeprecationWarning: XML declarations in HTML module descriptions are deprecated since Odoo 17, base_iso3166 can just have a UTF8 description with not need for a declaration.`

ramiadavid: But in some pre-commit modules it is adding it

pedrobaeza: Yes, but it's also removed on the XML file, not only on the HTML.

base_iso3166/views/country_view.xml
ramiadavid: It's possible, I'll try to delete it...

ramiadavid: Deleted

base_iso3166/static/description/index.html
ramiadavid: I added it again, check the tests, you will see the warining in "Run tests" section...

pedrobaeza: Yeah, I mean in the XML file, not in the HTML file. Anyway, if it's added by the readme generator, we need to adjust such generator to not add the XML header in this file. cc @sbidoul @yajo @legalsylvain

ramiadavid: Yes, I misunderstood, it should not be deleted in XML, only in HTML, I have already added it again

miquelalzanillas (approved):

base_iso3166/static/description/index.html
sbidoul: I think the readme generator will re-add it on merge. This is an xhtml file, so that should be ok.

ramiadavid: @sbidoul But odoo 17 says that it is deprecated, we would have to modify the readme generator so that it does not add it, otherwise the odoo warning log will be filled.

In this specific module the index.html file is not generated by the generator. In others yes and when executing pre-commit it adds it again.

sbidoul: > odoo 17 says that it is deprecated

@ramiadavid Can you point me the warning (or better the code in Odoo that emits that deprecation)?

> In this specific module the index.html file is not generated by the generator

How so?

ramiadavid: https://github.com/OCA/maintainer-tools/blob/f71041f22b8cd68cf7c77b73a14ca8d8cd190a60/tools/gen_addon_readme.py#L403

pedrobaeza: @ramiadavid that's not the Odoo core code emitting the error.

ramiadavid: I'm going to review it, but I doubt it, you just have to read the error DeprecationWarning: XML declarations in HTML module descriptions are deprecated since **Odoo 17**, base_iso3166 can just have a UTF8 description with not need for a declaration.

ramiadavid: @pedrobaeza @sbidoul 
https://github.com/odoo/odoo/blob/bbb4a2e08adc1c93a4d65344fec01b69c8e2a744/odoo/addons/base/models/ir_module.py#L170

pedrobaeza: Thanks, the README generator should be changed then. Meanwhile, I think we can continue with the warning.

pedrobaeza (approved):
/ocabot merge nobump

PR #183, opened 2023-12-29
[17.0][MIG] product_fao_fishing: Migration to 17.0

youelEl:


gurneyalex (approved):

pedrobaeza:
@gurneyalex please try to avoid to launch too many merge commands in the same branch for avoiding to repeat again and again the CIs. Merge one, and when done, launch the following. This branch is not very heavy (for now), but on other branches with a lot of modules, it struggles whole OCA queues.

PR #185, opened 2023-12-29
[16.0][FIX] uom_unece: fix unece code for kilometer

alexis-via:
Migration script to fix the unece code is provided. Re-order data/unece.xml to be consistent with uom/data/uom_data.xml
See comment in the XML file to know more about the cause of the initial error.
I found this bug while reading the new unece mapping of odoo v17 https://github.com/odoo/odoo/blob/17.0/addons/account/models/uom_uom.py#L20


rlemeur (approved):

pedrobaeza (changes requested):
No need of migration scripts, as they are noupdate=0.

Please also don't increase the diff "reordering" things, as they don't add value and complicate the review or later blame.

alexis-via:
@pedrobaeza We need a migration script because the data uom.uom added by the "uom" module is noupdate=1
https://github.com/odoo/odoo/blob/16.0/addons/uom/data/uom_data.xml#L2
The noupdate=0 in the module is meaning less. I'll remove it.

alexis-via:
@pedrobaeza I answered your remark. Could you remove your change request please ?

pedrobaeza (approved):
/ocabot merge minor

alexis-via:
Thanks Pedro !

PR #187, opened 2023-12-30
[FIX] [15.0] base_iso3166 breaking tests if code is False

ghost:
The module currently breaks tests in "delivery_multi_destination".

I added a test if code is False before getattr.

https://github.com/OCA/delivery-carrier/actions/runs/7362481660/job/20040960822?pr=751

Bildschirmfoto 2023-12-30 um 08 20 17

### DIFF
--- a/base_iso3166/models/res_country.py
+++ b/base_iso3166/models/res_country.py
@@ -27,7 +27,9 @@ class ResCountry(models.Model):
 
     @api.depends("code")
     def _compute_codes(self):
-        for country in self:
+        self.code_alpha3 = False
+        self.code_numeric = False
+        for country in self.filtered("code"):
             c = False
             for country_type in ["countries", "historic_countries"]:
                 try:
@@ -39,6 +41,3 @@ class ResCountry(models.Model):
             if c:
                 country.code_alpha3 = getattr(c, "alpha_3", getattr(c, "alpha3", False))
                 country.code_numeric = c.numeric
-            else:
-                country.code_alpha3 = False
-                country.code_numeric = False


base_iso3166/models/res_country.py
```diff
@@ -29,13 +33,15 @@ class ResCountry(models.Model):
     def _compute_codes(self):
         for country in self:
```
pedrobaeza:
```suggestion
        for country in self.filtered("code"):
```

pedrobaeza (changes requested):
Don't add logger. It's annoying. And reduce the diff simply filtering on the for.

base_iso3166/models/res_country.py
ghost: A) in the case of code=False, it's not calculating  "code_alpha3" and "code_numeric" which will produce errors
-> I would need to handle this case outside of the loop which would make the code more complex.
or do I miss something?

pedrobaeza: Do a previous `self.code_alpha3 = False`. And please squash everything in one commit.

pedrobaeza: You are still not implementing this and changing the existing code, so CI has broken

base_iso3166/models/res_country.py
```diff
@@ -27,18 +27,18 @@ class ResCountry(models.Model):
...
+        for country in self.filtered("code"):
             c = False
-            for country_type in ["countries", "historic_countries"]:
-                try:
-                    c = getattr(pycountry, country_type).get(alpha_2=country.code)
-                except KeyError:
-                    c = getattr(pycountry, country_type).get(alpha2=country.code)
-                if c:
-                    break
+            if country.code:
```
pedrobaeza: Remove this `if` and restore the previous code, and you still have several commits.

ghost:
now it should be a clean commit

base_iso3166/models/res_country.py
```diff
@@ -27,18 +27,15 @@ class ResCountry(models.Model):
...
-        for country in self:
+        self.code_alpha3 = False
+        self.code_numeric = False
+        for country in self.filtered("code"):
             c = False
             for country_type in ["countries", "historic_countries"]:
                 try:
                     c = getattr(pycountry, country_type).get(alpha_2=country.code)
                 except KeyError:
-                    c = getattr(pycountry, country_type).get(alpha2=country.code)
```
pedrobaeza: Why are you removing this part? It's for compatibility with old versions.

ghost: it's back

pedrobaeza: Incorrect indentation and the break is missing.

ghost: added the break, where is an incorrect indentation?

pedrobaeza (approved):
Finally it's correct!

Issue #129, opened 2022-02-17
Compute BIC automatically from IBAN


mschulz-asphericon:
regarding "base_bank_from_iban"

**Is your feature request related to a problem?**

Because the IBAN is sufficient to fetch the corresponding BIC and Country, it would be a great improvement to do this automatically.

Another problem is, that the module did not quite work for me, following the same steps as in the test instruction and the test file. For the exact same data it worked, but as soon as I have changed to German banks and partners it did not work correctly.

**Describe the solution you'd like**

On a partner's bank account list: when entering an IBAN it should check for available banks, if none is available it should suggest/ask if a new bank should be created with the information available from this API: 
https://openiban.com/validate/DE89370400440532013000?getBIC=true

**Describe alternatives you've considered**

**Additional context**

When creating SEPA mandates sometimes BIC codes are necessary, but these are not always available in the database and it is time consuming to check for the BIC codes, while this can be automated.

Looking forward to your response, best regards
Michael

bosd:
1.
Using it on dutch IBAN'S and it doesn't work for me either.

The module stores the `code` in a separate field. While the code is already present in the IBAN.
Therefore better to change code to a computed field.
Add some logic to extract the bank code from the iban for the matching.

@oca/maintainers can you place a bug label on this issue?

2.
Your proposed solution is to validate/lookup with an external api. 
Looks cool, but should be a separate module.

bosd:
Please reopen

pedrobaeza:
Do you have any implementation?

bosd:
> Do you have any implementation?

No, i don't have an emplementation.
Couple of days ago i stumbled upon this cool lib.


https://github.com/mdomke/schwifty

pedrobaeza:
That library seems to do the same of the `base_bank_from_iban` module.

bosd:
Yes, but the module is broken??
I don't remember correctly because it was some time ago.
I tested it manually against a bunch of different ibans from different countries.
After that ignored it, because it was not usable in my opinion. Similar to @mschulz-asphericon 

Ths does'nt mean the module does not work at all.. maybe against a very tiny subset of ibans.
(Which I did not find)

The earlier mentioned module can be an answer of importing bank accounts.
There was an discussion here, but i cannot find it again. 

About finding a good (open/free) data set for bank lists (with bic).
At the time of the discussion I could not find any good reliable source. even not at paid one. 
The problem is, that there is no official data set available. with good bank / bis lists.
[...]

PR #190, opened 2024-01-15
[14.0][FIX][IMP] base_bank_from_iban

bosd:
Fixes issue #129

Before this PR the module worked counter intuitive.
It relies on the field `code` which is introduced by the module itself.

Since it is not a native odoo field. It is not populated with data from one of the bank list modules:
e.g. https://github.com/OCA/l10n-croatia/tree/16.0/l10n_hr_bank
https://github.com/OCA/l10n-netherlands/tree/14.0/l10n_nl_bank

An option was to populate that field. But then all these localization modules needed to be adjusted.
AFAIK the field `code` has no real use case, then just to search. 
As opposed to a BIC field. Which could be used in payment (xml) files. 
And should be  visible on invoices. Source: https://www.ecbs.org/iban.htm

This adds support to match a bank based upon the BIC code which is encapsulated inside the IBAN.
Which is used by iban's from the following countries:
UK, Netherlands, Bulgaria, Malta, Latvia, Romania, Gibraltar

https://en.wikipedia.org/wiki/International_Bank_Account_Number#IBAN_formats_by_country

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,7 +1,8 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.17.2
+_commit: v1.20
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
+convert_readme_fragments_to_markdown: false
 generate_requirements_txt: true
 github_check_license: false
 github_ci_extra_env: {}
@@ -29,4 +30,6 @@ repo_description: 'This project is a collection of Odoo modules containing vario
 repo_name: Community Data Files
 repo_slug: community-data-files
 repo_website: https://github.com/OCA/community-data-files
+use_pyproject_toml: false
+use_ruff: false
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ __pycache__/
 *.py[cod]
 /.venv
 /.pytest_cache
+/.ruff_cache
 
 # C extensions
 *.so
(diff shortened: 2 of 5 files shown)


bosd:
@Tecnativa Can you have a look?

pedrobaeza:
Please don't ping to Tecnativa, as it's altering our stats.

pedrobaeza (changes requested):
Doing a `like` search is "likely" to fail, as it will match too many records. The idea of using `code` field is to extract the exact part of the IBAN that represents the matching.

This doesn't serve IMO.

bosd:
What would be an alternative way to search for the `bban` which is encapsulated inside the `BIC`?
As an exact match is not possible.



image
Source: https://www.swiftbic.com/what-is-bic-code.html




Example added from test in this pr:

`NL36 INGB 0003445588` results in BBAN code `INGB`. 
It should match the bank with BIC `INGBNL2A`
Combined with the filter on the bank's country.code it is likely to hit only one result.


From the formatting of the BIC we could add country code after the bban `code` to make it more resilliant.
`"bic", "like", bank_code+country_code.upper())`
result `INGBNL`
(But I don't see a way to add the last characters  a.k.a. Location code `2A` so we could exact match.)



Anyway, AFAIK
The BBB code is already designed to be unique.
E.g. from some examples in Netherlands... 
| Code   |  Bank name |
| ABNA   | ABN AMRO BANK |
| ABNC    | ABN AMRO CLEARING BANK N.V |

pedrobaeza:
For doing that, please create another module. Call it `base_bank_from_bic` or similar.

PR #181, opened 2023-12-21
17.0 mig base unece

youelEl:

PR #193, opened 2024-03-02
[17.0][MIG] base_unece

alexis-via:
This PR overrides https://github.com/OCA/community-data-files/pull/181 which is stalled


pedrobaeza (approved):
The _rec_names change is debatable, but I don't oppose to use it the way you intend.

PR #195, opened 2024-04-04
Improve migration 16.0 l10n eu nace

aronabencherif:
This is an improvement on the migration of the` l10n_eu_nace `module to version 16, taking into account feedback received on this PR [https://github.com/OCA/community-data-files/pull/176](https://github.com/OCA/community-data-files/pull/176)


edlopen:
Hello @aronabencherif , 
it seems that both of us worked on the same module. Here is [my PR](https://github.com/OCA/community-data-files/pull/196) and it uses the new ShowVoc endoint as RAMON is no longer updated. Also I ditched the res.partner.nace model because is not necessary having the res.partner.industry from Odoo as @yajo [said](https://github.com/OCA/community-data-files/pull/176#pullrequestreview-1749999335).

I would like to ask if you could have a look to the PR as is my intention to migrate and update other modules that relays on this new dataset cataloge ([ShowVoc](https://ec.europa.eu/eurostat/web/metadata/classifications)).

Thank you in advance.

PR #202, opened 2024-07-12
[IMP] uom_unece: Change unit code from C62 to PCE

Rad0van:
According to https://www.gs1us.org/resources/data-hub-help-center/unit-of-measure-codes **C62** means _one_ while **PCE** means _piece_. I think **PCE** is correct as for example _dozen_ is encoded as **DPC** meaning _dozen piece_.

### DIFF
--- a/uom_unece/data/unece.xml
+++ b/uom_unece/data/unece.xml
@@ -2,7 +2,7 @@
 <odoo>
     <!-- Units -->
     <record id="uom.product_uom_unit" model="uom.uom">
-        <field name="unece_code">C62</field>
+        <field name="unece_code">PCE</field>
     </record>
     <record id="uom.product_uom_dozen" model="uom.uom">
         <field name="unece_code">DPC</field>


pedrobaeza:
Please follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Rad0van:
> Please follow commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Thanks Pedro for reminder. Need  to remind this once for all.

uom_unece/data/unece.xml
```diff
@@ -2,7 +2,7 @@
 <odoo>
     <!-- Units -->
     <record id="uom.product_uom_unit" model="uom.uom">
-        <field name="unece_code">C62</field>
+        <field name="unece_code">PCE</field>
```
bosd, line 5:
```suggestion
        <field name="unece_code">H87</field>
```

bosd (changes requested):
Actually the PCE is the old code.
The new recomended code is `H87`.
See:
https://www.gs1.org/standards/eancom/eancom-2002-edition-2016/eancom-syntax-3
click on:
Part III: Data elements & code sets directory 
then go to:
[2. Index by data element name](https://www.gs1.org/sites/default/files/docs/eancom/ean02s3/part3/dn1.htm)
search for:
`Measurement unit code `

Then you will reach this page..

[Measurement unit code](https://www.gs1.org/sites/default/files/docs/eancom/ean02s3/part3/dc126.htm)

Where the recomended codes are explained.

image

bosd:
We need a migration script for this change as well.

alexis-via (changes requested):
No, in REAL LIFE usage, C62 is used AFAIK.

bosd:
Is there an option to support multiple codes for the same item?

As we can encounter the situation that we receive an document for 

partner A with code   H87
And partner b with C62

Both seems valid. 
As it may take a long time for partners to migrate from the "old" to "new" code.

alexis-via:
@bosd I agree with you, the real topic is "how can he have 1 code for generation (C62) and several codes for import (C62, PCE, H87)". I don't have great ideas on this at the moment, but it's definitely the problem to tackle.

bosd:
Could we leave this one open?

PR #206, opened 2024-09-28
[16.0][FIX] base_bank_from_iban: Handle correctly non IBAN accounts

pedrobaeza:
If any non IBAN account is provided, there's an ugly log with traceback each time, polluting tests and system logs:

```
INFO prod odoo.addons.base_bank_from_iban.models.res_partner_bank: Could not find bank from IBAN
Traceback (most recent call last):
  File ".../addons/base_bank_from_iban/models/res_partner_bank.py", line 34, in _add_bank_vals
    bank = self._get_bank_from_iban(vals["acc_number"])
  File ".../addons/base_bank_from_iban/models/res_partner_bank.py", line 42, in _get_bank_from_iban
    iban = schwifty.IBAN(acc_number)
  File ".../python/site-packages/schwifty/iban.py", line 77, in __init__
    self.validate(validate_bban)
  File ".../python/site-packages/schwifty/iban.py", line 175, in validate
    self._validate_characters()
  File ".../python/site-packages/schwifty/iban.py", line 185, in _validate_characters
    raise exceptions.InvalidStructure(f"Invalid characters in IBAN {self!s}")
schwifty.exceptions.InvalidStructure: Invalid characters in IBAN XXXXXX
```

This commit removes that traceback catching the proper exception, and handling it accordingly, and also removing an extra INFO log that was not adding value.

### DIFF
--- a/base_bank_from_iban/tests/test_base_bank_from_iban.py
+++ b/base_bank_from_iban/tests/test_base_bank_from_iban.py
@@ -1,5 +1,5 @@
 # Copyright 2017 Tecnativa - Carlos Dauden
-# Copyright 2022 Tecnativa - Pedro M. Baeza
+# Copyright 2022,2024 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3).
 
 from odoo.tests import Form, common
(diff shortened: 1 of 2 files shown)


Andreykine (commented):
T

base_bank_from_iban/tests/test_base_bank_from_iban.py
```diff
@@ -1,5 +1,5 @@
 # Copyright 2017 Tecnativa - Carlos Dauden
-# Copyright 2022 Tecnativa - Pedro M. Baeza
+# Copyright 2022,2024 Tecnativa - Pedro M. Baeza
```
carlosdauden, line 2: Is this needed or not?:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#tasks-not-to-do-in-the-migration

carlosdauden (approved):

base_bank_from_iban/tests/test_base_bank_from_iban.py
pedrobaeza: Well, I have adapted the format to not put as ending date, but "continuing" date. I want to highlight that it has been touched in these 2 years.

carlosdauden: So what should I put in a module that for example has been modified in the years: 2016, 2017, 2020 and 2023?

pedrobaeza: I would put 2016,2023

carlosdauden: And when would the usual 2016-2023 be used?

pedrobaeza: Well, @sbidoul put that note because he says that putting that notation means that the copyright ends on that date, which is right, so I have somehow invented this variant to highlight some work between several years, only putting the first and the last you have touched it.

PR #208, opened 2024-10-09
[18.0][MIG] l10n_eu_product_adr: Migration to 18.0

chaule97:
- I removed the test case `test_06_product_variant_migration` because there is no more pre-migration in Odoo 16+.
- I need to update the `_compute_display_name` method in `AdrClass` and `AdrGoods` because the `name_search` method in Odoo 18 has changed and use `display_name` value to return result.


l10n_eu_product_adr/views/product_template_views.xml
```diff
@@ -19,24 +19,18 @@
                     name="adr"
                     string="Dangerous Goods"
                     groups="l10n_eu_product_adr.group_adr_goods_user"
-                    attrs="{'invisible': ['|', ('type', '!=', 'product'), '&amp;', ('is_dangerous', '=', False), ('adr_goods_on_variants', '=', False)]}"
+                    invisible="not (is_storable and (is_dangerous or adr_goods_on_variants))"
```
mmequignon: are we sure this is the same ?
I would have written this `not is_storable or (is_dangerous and adr_goods_on_variants)`

chaule97: no, it will be: `not is_storable or (not is_dangerous and not adr_goods_on_variants)`

TDu (commented):
Testing on runboat I have an issue when activating the Is Dangerous flag on a product...the Dangerous Goods tab is empty.

l10n_eu_product_adr/views/product_template_views.xml
TDu: Thanks for the change it looks better IMO

chaule97:
> Testing on runboat I have an issue when activating the Is Dangerous flag on a product...the Dangerous Goods tab is empty.

Hi @TDu ,I have checked the differences between  [PR 16.0](https://github.com/OCA/community-data-files/blob/d2bed6305fec8fa673e4d96485f5a9e6e51378e1/l10n_eu_product_adr/views/product_product_views.xml) and [branch 14.0](https://github.com/OCA/community-data-files/blob/14.0/l10n_eu_product_adr/views/product_product_views.xml). The PR 16.0 had incorrect group permissions. I have updated. thank you

chaule97:
Excuse me, Why does pre-commit run correctly on my local environment?

TDu (approved):

l10n_eu_product_adr/tests/test_adr_goods.py
```diff
@@ -160,57 +158,3 @@ def test_05_product_variant_create(self):
...
             var.adr_goods_id == self.goods1 for var in template.product_variant_ids
         )
-
-    # def test_06_product_variant_migration(self):
-    #     """Test the migration of the adr fields from template to product"""
-    #     try:
-    #         from openupgradelib import openupgrade  # noqa: F401
-    #     except ImportError:
-    #         logging.getLogger("odoo.addons.l10n_eu_product_adr.tests").info(
-    #             "OpenUpgrade not found, skip test"
```
mmequignon: I wouldn't drop this test, as it probably had meaning in the past.
It is probably meant to be activated again once code is fixed.

mmequignon (approved):

l10n_eu_product_adr/tests/test_adr_goods.py
chaule97: Hi @mmequignon , I think this was the test that went with the migration code for version 14, it's obsolete/not needed now

chaule97:
Ciao @simahawk , could you merge again please?

PR #221, opened 2024-12-19
[18.0][MIG] account_payment_unece

alexis-via:
I'll publish account_payment_base_oca in OCA/bank-payment ASAP


account_payment_unece/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+{
+    "name": "Account Payment UNECE",
+    "version": "18.0.1.0.0",
+    "category": "Accounting & Finance",
+    "license": "AGPL-3",
+    "summary": "UNECE nomenclature for the payment methods",
+    "author": "Akretion,Odoo Community Association (OCA)",
+    "maintainers": ["alexis-via"],
+    "website": "https://github.com/OCA/community-data-files",
+    "depends": ["account_payment_base_oca", "base_unece"],
```
grindtildeath: Do we really need the dependency on `account_payment_base_oca`?

alexis-via: The dependency is ONLY for the views. There is no form/list view of account.payment.method in the "account" module.

alexis-via:
My first idea was to move the form/list view of account.payment.method to account_usability, so that the module can be used by users of OCA/bank-payment AND users of OCA/bank-payment-alternative (where account_payment_base_oca is located).
But I think users of Odoo Enterprise don't want to install account_usability.
So my second idea is to move the form/list view of account.payment.method to a small module account_payment_method_views ; both this module and account_payment_base_oca and account_payment_mode would depend on it.

alexis-via:
In the meantime: account_payment_base_oca is located in this PR: https://github.com/OCA/bank-payment-alternative/pull/3

alexis-via:
I propose to add the module account_payment_method_views to OCA/account-financial-tools.
If you disagree with this proposal, please make alternative proposals now !

alexis-via:
Here it is: https://github.com/OCA/account-financial-tools/pull/2094
The name I propose for this module is account_payment_method_base

account_payment_unece/__manifest__.py
alexis-via: Updated, see my comments below

alexis-via:
account_payment_method_base has been merged in OCA/account-payment, cf https://github.com/OCA/account-payment/pull/840

PR #222, opened 2025-01-21
[FIX][18.0] l10n_eu_product_adr: Fixed the kanban image in the adr label level.

bizzappdev:
Reason: There is an error when opening or clicking the tab "Dangerous Goods" in the product form view or similar views where Kaban view of the "adr.label" model has been added.

kanban-img-issue

### DIFF
--- a/l10n_eu_product_adr/views/adr_label_views.xml
+++ b/l10n_eu_product_adr/views/adr_label_views.xml
@@ -25,10 +25,13 @@
                             t-attf-class="oe_kanban_global_click"
                             style="min-width:450px;min-height:80px"
                         >
-                            <div class="o_kanban_image">
-                                <img
-                                    alt="Label image"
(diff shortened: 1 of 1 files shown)


l10n_eu_product_adr/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "ADR Dangerous Goods",
     "summary": "Allows to set appropriate danger class and components",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
```
simahawk: there's no need to change the version of the module. The oca-bot will take care ;)

simahawk (changes requested):
Thanks for your help!

As you are fixing up, can you please shorten the commit msg?

`[FIX] l10n_eu_product_adr: fix adr label kanban view`

Sounds cleaner ;) 
Thanks!

bizzappdev:
@simahawk, Both the changes are done

PR #230, opened 2025-02-19
[18.0][MIG] product_fao_fishing: Migration to 18.0

CarlosRoca13:
This PR depends on:

- [x]  https://github.com/OCA/product-attribute/pull/1892

cc @Tecnativa TT54226

ping @carlosdauden @sergio-teruel


carlosdauden:
I believe that the new module product_attribute_value_auto_fill_option should be added as a dependency and that we should set the fill_all_values value to True for the attributes created in this module.

CarlosRoca13:
> I believe that the new module product_attribute_value_auto_fill_option should be added as a dependency and that we should set the fill_all_values value to True for the attributes created in this module.

I think that it should not be set as true to avoid select all the values by default... isn't it?

CarlosRoca13:
Added dependency, the value of the field will be true by default so no need migration scripts

carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza:
You can rebase removing the dependency

CarlosRoca13:
Done @pedrobaeza

PR #235, opened 2025-04-29
[17.0][IMP] base_bank_from_iban: raise clean error for iban validation

kevinkhao:
This module will raise an ugly exception with stacktrace. Now just raise a validation error with a more user-friendly message


pedrobaeza (commented):
Please reduce the noise in the diff not adding the contributor in the README for such simple 2 lines change, please. You can put your copyright in the same file though.

PR #237, opened 2025-05-05
[ADD][18.0] base_iso7010_data_mandatory

bosd:
Depends on #236 
This actually adds the Safety Symbols to use on other odoo models and reports.


bosd:
please keep open

PR #243, opened 2025-09-19
[15.0][IMP] company_sanitary_registry: Add model to manage multiple sanitary registries and add it to the product categories

CarlosRoca13:
cc @Tecnativa TT58084

ping @sergio-teruel @carlosdauden

### DIFF
--- a/company_sanitary_registry/models/res_company.py
+++ b/company_sanitary_registry/models/res_company.py
@@ -6,4 +6,7 @@ from odoo import fields, models
 class ResCompany(models.Model):
     _inherit = "res.company"
 
-    sanitary_registry = fields.Char()
+    sanitary_registry_id = fields.Many2one("sanitary.registry")
+    # Field to avoid crashes modules that uses yet this field. To be removed
+    # On future versions.
+    sanitary_registry = fields.Char(related="sanitary_registry_id.name")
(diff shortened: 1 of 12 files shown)


company_sanitary_registry/models/res_company.py
```diff
@@ -6,4 +6,4 @@
 class ResCompany(models.Model):
     _inherit = "res.company"
 
-    sanitary_registry = fields.Char()
```
sergio-teruel, line 9: What do you think about maintain this field as related to first sanitary registry to minimize initial fails in other modules that still uses the old field?

company_sanitary_registry/models/sanitary_registry.py
```diff
@@ -0,0 +1,25 @@
...
+class SanitaryRegistry(models.Model):
+    _name = "sanitary.registry"
+    _description = "Sanitary Registry"
+
+    name = fields.Char()
+
+    @api.constrains("name")
+    def _check_name(self):
+        for san_reg in self:
+            other_reg = self.env["sanitary.registry"].search(
```
sergio-teruel: Can you do the search outside of loop??

sergio-teruel (changes requested):

CarlosRoca13:
Done @sergio-teruel

carlosdauden (approved):

carlosdauden:
ping @sergio-teruel

sergio-teruel (approved):

PR #250, opened 2025-12-02
[17.0][IMP] l10n_eu_product_adr

edescalona:
@BinhexTeam

## What

`stock.move.adr_points` used to resolve the points factor during every recalculation: tracing from the stock move to its product, via `adr_goods_id` and then through `un_number_points_factor_map`, using  category_points_factor_map` as a fallback. Since the factor depends solely on the ADR goods record, it is now stored once on the product itself.

## Changes

- New field `product.product.adr_factor`, calculated from `adr_goods_id` (`store=True`, `precompute=True`, `readonly=False` to allow overrides at the product level).
- `stock.move._compute_adr_points` reads this field instead of resolving the maps itself. The factor maps have been moved out of `stock_move.py`.
- A `pre_init_hook` populates the column in a single SQL pass during installation; likewise, a `migrations/17.0.1.1.0/pre-migration.py` file performs the same operation for existing databases. The column is added with a default value of 0 (an instant operation that does not rewrite the table), and only products with an `adr_goods_id` are updated; this ensures the installation remains viable for large catalogs where a record-by-record ORM calculation would not be feasible.
- Version updated to `17.0.1.1.0`.

## Behavior

The SQL query exactly replicates the logic of `_compute_adr_factor`, including the fallback to a value of 0; consequently, products without dangerous goods—as well as those with transport categories not listed in the factor map (`0`, `-`, `CARRIAGE PROHIBITED`, `Not subject to ADR`)—receive a value of 0, just as before. The `adr_points` values ​​remain unchanged.

### DIFF
--- a/l10n_eu_product_adr/i18n/es.po
+++ b/l10n_eu_product_adr/i18n/es.po
@@ -13181,3 +13181,8 @@ msgstr ""
 #: model:adr.goods,name:l10n_eu_product_adr.adr_goods_2747
 msgid "tert-BUTYLCYCLOHEXYL CHLOROFORMATE"
 msgstr ""
+
+#. module: l10n_eu_product_adr
+#: model:ir.model.fields,field_description:l10n_eu_product_adr.field_product_product__adr_factor
+msgid "Adr Factor"
+msgstr "Coeficiente Adr"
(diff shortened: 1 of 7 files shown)


edescalona:
ping @rrebollo

l10n_eu_product_adr/i18n/es.po
```diff
@@ -13181,3 +13181,8 @@ msgstr ""
 #: model:adr.goods,name:l10n_eu_product_adr.adr_goods_2747
 msgid "tert-BUTYLCYCLOHEXYL CHLOROFORMATE"
 msgstr ""
+
+#. module: l10n_eu_product_adr
+#: model:ir.model.fields,field_description:l10n_eu_product_adr.field_product_product__adr_factor
+msgid "Adr Factor"
+msgstr "Factor Adr"
```
rrebollo:
```suggestion
msgstr "Coeficiente Adr"
```

rrebollo (approved):
Code Review. LGTM! This improvement aims to make the process of the computation of the ADR points more transparent and traceable.

rrebollo (approved):
Code Review. LGTM!

Victor-Plagaro (approved):
Functional review OK, LGTM

PR #256, opened 2025-12-09
[MIG][19.0] account_payment_unece

alexis-via:


alexis-via:
Requires https://github.com/OCA/community-data-files/pull/255

PR #261, opened 2026-02-18
[16.0][FIX] base_bank_from_iban: Add new except to _get_bank_from_iban

Andrii9090-tecnativa:
To avoid errors, we must add exceptions to the `_get_bank_from_iban` method of the `res.partner.bank` model.
```
File "/opt/odoo/auto/addons/account_payment_partner/tests/test_account_payment_partner.py", line 523, in test_refund_no_payment_mode_preserves_partner_bank
    trusted_bank = self.env["res.partner.bank"].create(
  File "<decorator-gen-782>", line 2, in create
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 430, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/auto/addons/base_bank_from_iban/models/res_partner_bank.py", line 21, in create
    vals_list = [self._add_bank_vals(vals) for vals in vals_list]
  File "/opt/odoo/auto/addons/base_bank_from_iban/models/res_partner_bank.py", line 21, in <listcomp>
    vals_list = [self._add_bank_vals(vals) for vals in vals_list]
  File "/opt/odoo/auto/addons/base_bank_from_iban/models/res_partner_bank.py", line 30, in _add_bank_vals
    vals["bank_id"] = self._get_bank_from_iban(vals["acc_number"]).id
  File "/opt/odoo/auto/addons/base_bank_from_iban/models/res_partner_bank.py", line 36, in _get_bank_from_iban
    iban = schwifty.IBAN(acc_number)
  File "/usr/local/lib/python3.10/site-packages/schwifty/iban.py", line 77, in __init__
    self.validate(validate_bban)
  File "/usr/local/lib/python3.10/site-packages/schwifty/iban.py", line 178, in validate
    self._validate_iban_checksum()
  File "/usr/local/lib/python3.10/site-packages/schwifty/iban.py", line 202, in _validate_iban_checksum
    raise exceptions.InvalidChecksumDigits("Invalid checksum digits")
schwifty.exceptions.InvalidChecksumDigits: Invalid checksum digits
```
@pedrobaeza can you review?
@Tecnativa

### DIFF
--- a/base_bank_from_iban/models/res_partner_bank.py
+++ b/base_bank_from_iban/models/res_partner_bank.py
@@ -56,7 +56,10 @@ class ResPartnerBank(models.Model):
                     bank = self.env["res.bank"].create(vals)
             else:
                 bank = self.env["res.bank"]
-        except schwifty.exceptions.InvalidStructure:
+        except (
+            schwifty.exceptions.InvalidStructure,
+            schwifty.exceptions.InvalidChecksumDigits,
+        ):
             bank = self.env["res.bank"]
         return bank
 


pedrobaeza (approved):
Thank you.

Fast-tracking this obvious one:

/ocabot merge patch

Please fw-port it

qgroulard:
Hi @pedrobaeza @Andrii9090-tecnativa, this is not so obvious in my opinion.

There is a key difference between `InvalidStructure` and `InvalidChecksumDigits`:
- `InvalidStructure`: The account number is not an IBAN at all;
- `InvalidChecksumDigits`: The account number is an IBAN, but a wrong one. In the same way as `InvalidLength` for instance.

Hence, before this PR, `_get_bank_from_iban` went silent when the account number was not an IBAN, but raised if it got a wrong one.
Now, after this PR, `_get_bank_from_iban` goes silent if the account number is not an IBAN or is a wrong one.

IMO, this is not a small fix but a real breaking change.

Personally I think this is a good shift as this is not the responsibility of this module to validate the IBAN format.
However:

1. The impacts of this change must not be underestimated;
2. We should not only consider `InvalidChecksumDigits`, but every formatting issue (see https://github.com/OCA/community-data-files/pull/288).

pedrobaeza:
This patch goal is to avoid crash on case of wrong IBAN, because this modules shouldn't handle that, so I don't get the problem.

PR #266, opened 2026-03-05
[18.0][FWPORT] company_sanitary_registry: Forward port non-migrated commits

CarlosRoca13:
cc @Tecnativa

ping @carlosdauden @sergio-teruel 

I’m not sure whether there will be any issues with the migration scripts when moving from a previous version to this one, but they are necessary in case the module is updated while already installed in 18.0.


company_sanitary_registry/migrations/18.0.1.1.0/pre_migration.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2025 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
```
carlosdauden, line 9: I think you should check whether the table exists and has records to avoid executing this code.

carlosdauden (changes requested):

company_sanitary_registry/migrations/18.0.1.1.0/pre_migration.py
```diff
@@ -0,0 +1,47 @@
...
+        env.cr,
+        """
+        INSERT INTO sanitary_registry (name)
+        SELECT sanitary_registry
+        FROM res_company
+        WHERE sanitary_registry IS NOT NULL
+            AND NOT EXISTS (
+              SELECT 1 FROM sanitary_registry LIMIT 1
+            )
```
CarlosRoca13, line 29: By adding this code to this sql I think that we will have no problem with the migration from older versions

carlosdauden (approved):

company_sanitary_registry/migrations/18.0.1.1.0/pre_migration.py
carlosdauden: OK

chienandalu (approved):
Thanks @CarlosRoca13 🙂 👍

PR #275, opened 2026-05-12
[19.0][ADD] uom_preset

bosd:
## Summary

Adds `uom_preset`, an abstract-model registry that translates supplier-specific UoM codes to UN/CEFACT Recommendation 20 codes (via `uom_unece`), so they resolve to `uom.uom` records.

The module is intentionally narrow in scope: it only covers nomenclatures that **do not** use UN/CEFACT codes natively. For SCSN-style or any other Rec 20-conformant supplier flow, callers should look up `uom.uom.unece_code` directly via `uom_unece` — no preset is needed.

## Vocabularies shipped

* `pricat` (11 entries) — the Dutch Pricat catalogue/order exchange format, sourced verbatim from INDI's public OCI-PunchOut documentation: https://www.indi.nl/nl-nl/slim-inkopen/erp-connecties/oci-punchout — section *Eenheden-overzicht*.
* `x12_355` (160 entries) — ANSI ASC X12 element 355, used in US retail and manufacturing EDI. Only the entries where the X12 code differs from the UN/CEFACT Rec 20 code are shipped; pass-through entries are dropped so callers fall through to `uom_unece` directly. Source: *Ariba Network UOM Mapping for ANSI X12* (SAP, 2014, [PDF](https://help.sap.com/doc/ba35b82316194c94961cfabb0284e988/cloud/en-US/uomMappingForANSI_X12Documents.pdf)).

## API

```python
uom = self.env["uom.preset"]._resolve("pricat", "STU")
# -> uom.uom record whose unece_code == "C62", or empty recordset on miss.
```

Resolution is strict: an unknown vocabulary or a code missing from the chosen vocabulary returns an empty recordset. The caller chains its own fallback.

To add a vocabulary in a follow-up module, inherit `uom.preset` and extend `_uom_preset_vocabularies` / `_uom_preset_selection` (see USAGE).

## Tests

16 cases including a guard that fails if anyone adds a pass-through entry to either dict (catches the "this vocab is just `uom_unece` again" mistake before review).


uom_preset/models/uom_preset.py
```diff
@@ -0,0 +1,251 @@
...
+        Callers handle their own fallback chain (UN/CEFACT lookup, name
+        match, default UoM, …).
+        """
+        UomUom = self.env["uom.uom"]
+        if not (vocabulary and supplier_code):
+            return UomUom.browse()
+        vocab = self._uom_preset_vocabularies().get(vocabulary)
+        if not vocab:
+            return UomUom.browse()
+        unece_code = vocab.get(supplier_code.strip().upper())
```
rrebollo:
```suggestion
        if vocabulary and supplier_code:
            vocab = self._uom_preset_vocabularies().get(vocabulary)
            if vocab:
                unece_code = vocab.get(supplier_code.strip().upper())
```

rrebollo (approved):
Code Review. LGTM!

Please consider my suggestions.
Would you be so kind to review my #290 in return?

uom_preset/models/uom_preset.py
bosd: Applied in 6e90557 — thanks. Since the suggestion drops the early returns, I initialise `unece_code = None` before the block so the miss path still returns an empty recordset:

```python
UomUom = self.env["uom.uom"]
unece_code = None
if vocabulary and supplier_code:
    vocab = self._uom_preset_vocabularies().get(vocabulary)
    if vocab:
        unece_code = vocab.get(supplier_code.strip().upper())
if not unece_code:
    return UomUom.browse()
return UomUom.search([("unece_code", "=", unece_code)], limit=1)
```

Behaviour is unchanged; the existing tests cover the unknown-vocabulary and unknown-code paths.

PR #277, opened 2026-07-05
[18.0][IMP] account_tax_unece: add VAT exemption reason codes (VATEX)

alexis-via:


account_tax_unece/data/unece_tax_vatex.xml
```diff
@@ -0,0 +1,807 @@
...
+            name="description"
+        >Supports EN 16931-1 rule BR-O-10. Only use with VAT category code O.</field>
+    </record>
+    <record id="tax_vatex_fr_franchise" model="unece.code.list">
+        <field name="type">tax_vatex</field>
+        <field name="code">VATEX-FR-FRANCHISE</field>
+        <field name="name">France domestic VAT franchise in base</field>
+        <field
+            name="description"
+        >VAT exemption for Micro companies when Revenue is lower than a threashold. For domestic invoicing in France.</field>
```
remi-filament:
```suggestion
        >VAT exemption for Micro companies when Revenue is lower than a threshold. For domestic invoicing in France.</field>
```

account_tax_unece/data/unece_tax_vatex.xml
```diff
@@ -0,0 +1,807 @@
...
+        >Exempt based on 2° of article 261 D of the Code Général des Impôts (CGI ; General tax code)</field>
+        <field name="description">Only for domestic invoicing in France.</field>
+    </record>
+    <record id="tax_vatex_fr_cgi261d_3" model="unece.code.list">
+        <field name="type">tax_vatex</field>
+        <field name="code">VATEX-FR-CGI261D-3</field>
+        <field
+            name="name"
+        >Exempt based on 3° of article 261 D of the Code Général des Impôts (CGI ; General tax code)
+Exonération de TVA - Article 261 D-3° du Code Général des Impôts</field>
```
remi-filament: Here is the only place with text in French, is it on purpose ?

remi-filament (changes requested):
Thanks @alexis-via 
2 minor comments for 2 typos.

I am also wondering whether you should use invoice_legal_notes from account.tax to display the name of the tax on the invoice PDF report ?
It is already computed on moves based on taxes here : https://github.com/odoo/odoo/blob/e37f0420450ed2c8f465305b19cbbcfa78162666/addons/account/models/account_move.py#L2172 and then displayed on invoices PDF report : https://github.com/odoo/odoo/blob/e37f0420450ed2c8f465305b19cbbcfa78162666/addons/account/views/report_invoice.xml#L260

Maybe you should make this field on account.tax a related to unece_vatex_id.name ? (I do not think it makes sense to fill a different value between this field and vatex code)

account_tax_unece/migrations/18.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    all_taxes = env["account.tax"].with_context(active_test=False).search([])
+    # _compute_unece_vatex_id() uses the VATEX codes identified by XMLID
+    # So we must re-run the compute method after the creation of the VATEX codes
+    all_taxes._compute_unece_vatex_id()
```
remi-filament, line 13: Hi @alexis-via you should probably pre-create the column in a pre-migration script then so that it is not run twice !

account_tax_unece/data/unece_tax_vatex.xml
alexis-via: fixed

account_tax_unece/data/unece_tax_vatex.xml
alexis-via: fixed

account_tax_unece/migrations/18.0.2.0.0/post-migration.py
alexis-via: pre-creating the column in a pre-mig script won't solve the problem because the required data (account_tax_unece.tax_vatex_eu_ic and account_tax_unece.tax_vatex_eu_g) will not be created yet when premig will be executed.

remi-filament: Yes, but in order to avoid double compute (once when Odoo creates the field and then on post-migration).
Usually in these cases on OpenUpgrade, we initialize the fields in DB (with empty or default True/False value) so that when Odoo creates them they already exist and it does not try to compute them.
Then in post-migration we compute them properly

PR #279, opened 2026-07-07
[16.0][MIG]: l10n eu product adr: Migration to 16.0

remyakrishnann:
Module Migration to 16.0


remyakrishnann:
Hi everyone! 👋

This PR is ready for review. If anyone is available, I'd really appreciate a review.

I'm planning to migrate l10n_eu_adr_report, which depends on this module, so getting this PR reviewed would help me proceed with the next migration. Thanks in advance!

remyakrishnann:
Hi @pedrobaeza  @rrebollo  👋

Could you please take a look at this PR when you get a chance?

This is the 16.0 migration of l10n_eu_product_adr, and it has been waiting for review for the past couple of months. Since you have both been involved with this module/migration before, I thought you might be the best people to ask for a review.

This migration is also a dependency for the l10n_eu_adr_report migration.

Please let me know if anything needs to be changed or addressed on my side. I’d really appreciate your review. 
Thanks a lot🙏

rrebollo:
Just a detail, the PR's title should be [16.0][MIG] l10n eu product adr: migration to 16.0

l10n_eu_product_adr/models/adr_class.py
```diff
@@ -0,0 +1,35 @@
...
+class AdrClass(models.Model):
+    _name = "adr.class"
+    _description = "Dangerous Goods Class"
+    _order = "code, name"
+
+    code = fields.Char(required=True)
+    name = fields.Char(required=True, translate=True)
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
```
rrebollo, line 16: Optional (16.0 guide): the guide notes you *may* use the new `_rec_names_search` class attribute to expose the fields to search for, without keeping the `name_search` override. This is not required — `name_search`/`name_get` are still valid in 16.0 (they change from 17.0) — so the current code is fine; it is only a possible simplification. If you adopt it, `adr.class` could declare `_rec_names_search = ["code", "name"]` (and `adr.goods` `["un_number", "name"]`), which may let you drop the override.

l10n_eu_product_adr/models/adr_goods.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def name_get(self):
+        """Format the class name"""
+        res = []
+        for rec in self:
+            name = "{} {}".format(rec.un_number, rec.name)
+            affixes = []
+            if rec.transport_category != "-":
+                affixes.append(_("cat:%s", rec.transport_category))
+            if rec.limited_quantity or rec.limited_quantity_uom_id:
```
rrebollo, line 98: This condition changed vs 14.0 and diverges from the merged **17.0** branch. 14.0 and 17.0 both use `if rec.limited_quantity:`, while here it is `if rec.limited_quantity or rec.limited_quantity_uom_id:` and the value is formatted as `limited_quantity or 0.0`. Effect: when only the UoM is set (no quantity), the display name becomes `qty:0.0 <uom>`. If that is intentional, fine; otherwise it is safer to keep 14.0's condition to avoid a behavior change the migration does not require.

l10n_eu_product_adr/models/adr_goods.py
```diff
@@ -0,0 +1,108 @@
...
+
+from .common import TRANSPORT_CATEGORIES, TUNNEL_RESTRICTION_CODES
+
+
+class AdrGoods(models.Model):
+    _name = "adr.goods"
+    _description = "Dangerous Goods"
+
+    active = fields.Boolean(default=True)
+    un_number = fields.Char(
```
rrebollo, line 15: Minor: the explicit `string="UN Number"` is redundant on this `Char` (the field name already produces the same label). Cosmetic only.

l10n_eu_product_adr/models/adr_label.py
```diff
@@ -0,0 +1,37 @@
...
+    image = fields.Binary(required=True)
+    class_id = fields.Many2one(comodel_name="adr.class", string="Class", required=True)
+    goods_ids = fields.Many2many(
+        comodel_name="adr.goods",
+        copy=False,
+        string="ADR Goods",
+        help="The dangerous goods to which this label is applied",
+    )
+
+    def unlink(self):
```
rrebollo, line 24: Minor (15.0 guide): prefer `@api.ondelete(at_uninstall=False)` over overriding `unlink()`. Optional — the merged 17.0 branch still overrides `unlink`, so keeping it is acceptable too.

l10n_eu_product_adr/models/product_template.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    adr_tunnel_restriction_code = fields.Selection(
+        related="product_variant_ids.adr_goods_id.tunnel_restriction_code",
+    )
+
+    @api.depends("product_variant_ids.adr_goods_id")
+    def _compute_adr_goods_on_variants(self):
+        for template in self:
+            template.adr_goods_on_variants = not all(
+                product.adr_goods_id == (template.product_variant_ids[0].adr_goods_id)
```
rrebollo, line 54: Consider `product_variant_ids[:1]` instead of `product_variant_ids[0]`. A template normally always has at least one variant, but templates with dynamic attributes (`create_variant == 'dynamic'`) can have zero variants; in that case `[0]` raises `IndexError`. The core uses the `[:1]` slice in the same pattern (e.g. the computed `product_variant_id`), so the slice is the safe form.

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

rrebollo:
@remyakrishnann, I have several ongoing PRs of my own. I'd be grateful if you could review some of them in return.

remyakrishnann:
@rrebollo Thank you for the review and Sure !!  I’ll go through your PRs and review them once I get some time. I appreciate the offer to review each other’s work.

rrebollo:
@remyakrishnann I'm confused. I had already approved the PR and only made some optional suggestions. You didn't change anything, yet you requested my review again… so, what's the situation?

remyakrishnann:
@rrebollo Sorry, I accidentally re-requested your review. I realized you had already approved the PR , so please ignore the review request — no changes were made to the PR. Apologies for the confusion!

l10n_eu_product_adr/models/adr_label.py
remyakrishnann: Since 17.0 also keeps unlink() overridden this exact way (confirmed above), I'd leave it as-is for consistency with the merged upstream branch rather than diverging to @api.ondelete(at_uninstall=False)

l10n_eu_product_adr/models/adr_goods.py
remyakrishnann: str.title() only capitalizes the first letter of each word, it doesn't know un is an acronym for United Nations. So string="UN Number" is not redundant; it's there specifically to get the acronym capitalized correctly. Both 14.0 and 17.0 keep this explicit string too

l10n_eu_product_adr/models/product_template.py
remyakrishnann: product_variant_ids[1:] is evaluated eagerly to build the outer iteration, but the body (which contains [0]) is lazily evaluated only when all() pulls an item. If product_variant_ids has 0 or 1 elements, [1:] is empty, the generator never advances, and [0] is never touched — all() short-circuits to True on the empty iterable without ever accessing index 0. I verified this with a quick instrumented test: for an empty recordset, only the slice [1:] gets called, never the index [0].

So [0] only executes when the loop actually has ≥1 item from [1:], which means product_variant_ids has ≥2 elements — index 0 is always guaranteed to exist in that branch. Dynamic-attribute templates with zero variants don't trigger an IndexError here.

Given that, switching to [:1] wouldn't fix a real bug in this spot — it'd be defensive code for a case that structurally can't happen. I'd leave it as-is rather than adopt the slice pattern here

l10n_eu_product_adr/models/adr_class.py
remyakrishnann: adr.goods: the override just does search([("un_number", operator, name)]), passing through whatever operator the caller gave. That's exactly what _rec_names_search = ["un_number", "name"] does for free. Safe to drop, but optional.
adr.class: the override hardcodes an exact "=" match on code regardless of the caller's operator, while name still falls back to fuzzy ilike via super(). _rec_names_search applies the same operator to every listed field, so adopting it here would make code search fuzzy too — e.g. typing "1" would match any code containing "1", not just code "1". That's a real behavior change, not a pure simplification.
So i would keep as it is

l10n_eu_product_adr/models/adr_goods.py
remyakrishnann: Fixed — reverted to the 14.0 condition (if rec.limited_quantity:), confirmed it matches the merged 17.0 branch too

rrebollo (approved):
Code Review. LGTM!You already had my aprove, why didU asked a

PR #282, opened 2026-07-27
[18.0][IMP] account_tax_unece: hide VAT code on purchase taxes

alexis-via:


alexis-via:
After some thoughts, I eventually decided to hide completely VATEX code on purchase taxes, to make things really clear for users and avoid to have some users that may want to create different purchase taxes for each VATEX code on purchase.

njeudy (approved):
ok for me !

bguillot (approved):
LGTM

PR #241, opened 2025-09-01
[18.0][IMP] l10n_eu_product_adr: migrate xml data files to csv

rrebollo:
Converting all data to CSV for faster processing and smaller file sizes. This was already implemented during the migration to v17.

@BinhexTeam

PR #290, opened 2026-08-31
[19.0][IMP] l10n_eu_product_adr: migrate xml data files to csv

rrebollo:
Adapted port of #241 to 19.0 — not a pure port.

**Genesis**
Porting `l10n_eu_product_adr` from 18.0 (XML) to 19.0 (CSV) as in #241, but 19.0 core already provides `uom.product_uom_milliliter` and the CSV port would have introduced a duplicate `data/uom.uom.csv` that cannot load on 19 (dropped `category_id`/`factor`/`uom_type`). Also the CSV conversion loses label pictograms and V19 grid overflows long ADR names.

**What this PR does vs #241 (pure XML→CSV):**
* **Keeps `data/adr_label.xml` as XML** (22 records) instead of CSV. CSV cannot express `type="base64" file="..."`, so `data/adr.label.csv` left `image` empty and `static/img/*.svg` unreferenced (`image required=True`). Fix also needed on 17.0/18.0.
* **Does not introduce `data/uom.uom.csv`** at all. `origin/19.0` already points all 240 `adr.goods` rows to core `uom.product_uom_milliliter`; not adding the duplicate keeps the branch installable.
* **Fixes name overflow in form** — V19 `sheet>group colspan=4 col=4` with `field colspan=4` overflows for long ADR names. Uses idiomatic Odoo pattern (`group` default + `field colspan="2" class="text-break"`, as in `hr_leave_views.xml:406`, `product_views.xml:137`) instead of table-based colspan.

**Commits (3 on top of origin/19.0):**
* `[IMP] l10n_eu_product_adr: migrate xml data files to csv` (adapted, without `uom.uom.csv`, 240 refs to core)
* `[FIX] l10n_eu_product_adr: restore label pictograms`
* `[FIX] l10n_eu_product_adr: fix name overflow in form`

**Verification**
* `grep -rn mililiter` → 0, `grep -c uom.product_uom_milliliter data/adr.goods.csv` → 240
* `grep -c file= data/adr_label.xml` → 22
* Form overflow fixed via `class="text-break"` (no custom CSS)
* `pre-commit run --all-files` → Passed
* `invoke test --cur-file l10n_eu_product_adr/tests/test_adr_goods.py` → 10 green
* No `README.rst`/`static/description/index.html` changes

@BinhexTeam T12289


bosd:
Thanks for the port! The data conversion itself checks out, I diffed the generated CSVs against the 19.0 XML record by record (2928 `adr.goods` + 16 classes + 22 labels + 116 packing instructions, all fields and m2m ordering) and they are identical. Two things before this goes in.

### 1. The label pictograms are lost

`data/adr_label.xml` loads each image from disk:

```xml
<field name="image" type="base64" file="l10n_eu_product_adr/static/img/ADR_1.svg"/>
```

CSV can't express `type="base64" file="..."`, so `data/adr.label.csv` has an **empty `image` column for all 22 rows**, and the SVGs in `static/img/` are now unreferenced. `image` is `required=True` and it's what the form/kanban view renders, so every hazard label ends up without its pictogram.

Could you keep `adr_label.xml` as XML? It's only 22 records, so CSV buys little there, and it's the only file using the `file=` construct. Heads-up: this already landed the same way on 17.0 and 18.0, so those branches need the same fix.

### 2. Don't let the duplicate UoM enter 19.0 in the first place

19.0 never had it: `origin/19.0`'s `adr_goods.xml` already points all 240 rows at core `uom.product_uom_milliliter` and ships no custom UoM. The duplicate is introduced by the CSV commit itself, carried over from the 17.0/18.0 CSVs.

It also can't load on 19, core dropped `category_id`, `factor`, `uom_type` and the `uom.product_uom_categ_vol` xmlid (uom now uses `relative_factor`/`relative_uom_id`), so the `data/uom.uom.csv` added in the first commit leaves the tree uninstallable until the second commit removes it.

Simplest fix: **don't add `data/uom.uom.csv` to 19.0 at all**. The `[FIX]` and `[REF]` commits then disappear and the diff is just the XML→CSV port.

That's safe for upgrades, too. On 18→19 the `adr.goods` reload repoints all 240 rows to core `ml` before Odoo's end-of-update cleanup drops the now-obsolete `l10n_eu_product_adr.product_uom_mililiter`; if someone has that UoM on their own products the delete just fails harmlessly and leaves an unused record. That's exactly what 19.0 does today, with no migration script.

So the migration is optional. If you'd like to keep it to also repoint user-created `adr.goods` rows, that's fine, but drop the `product_uom_milliliter` (double-`l`) entry from it, since that xmlid has never existed in any released version.

Minor: `readme/CONTRIBUTORS.md` and `readme/HISTORY.md` weren't updated, and it would help to note in the description that this PR carries the 19.0 port of #241.

bosd:
Also discovered another nit, while comparing V14 vs V19. Not related to your work, worth fixing.

V14:
image
V19
image

bosd:
:+1:  Almost there. Still a overflow but images are back.
image

rrebollo:
@bosd I think is good to go now.

PR #296, opened 2026-09-18
[18.0][IMP] account_tax_unece: improve constrains and VATEX mapping

alexis-via:
Add char field unece_vat_exemption_reason
Add python constrain to check UNECE config
Set field unece_categ_id as required on sale VAT taxes


alexis-via:
My latest findings made me have some strong doubts about whether we should always have a BT-121 (rule G1.40+G1.41) or if BT-121 is optional when BT-120 is set (rule BR-E-10, BR-AE-10, BR-G-10, BR-IC-10, BR-O-10, BR-AF-10, BR-AG-10).

When you look at the schematron Flux10, they want both BT-121 and BT-120 (https://github.com/fnfempe/France_RFE/blob/main/FNFE_RFE_INVOICE/Flux10/Flux10.sch#L504) but when you look at the schematron Factur-X CTC, you can have either BT-120 OR BT-121 cf https://github.com/fnfempe/France_RFE/blob/main/FNFE_RFE_INVOICE/Factur-X/EN16931/schematron/FACTUR-X_EN16931.sch#L506

Until we make things clear on this topic, I pause this PR.

alexis-via:
I found the answer to my question (cf BR-FR-MAP-27/28). So it allows me to avoid adding a new field (better !) ; so this PR is now just a small improvement on the existing code.

PR #297, opened 2026-09-21
[18.0][FIX] base_bank_from_iban: catch base SchwiftyException instead of an incomplete whitelist

sergio-teruel:
_get_bank_from_iban already needed a second except added on top of the original InvalidStructure (9ccc044, InvalidChecksumDigits), and now a real-world malformed account number raises InvalidLength too. All of these inherit from schwifty.exceptions.SchwiftyException, so catch that instead of whack-a-moling individual subclasses as new malformed-input shapes turn up.

@Tecnativa TT64436

ping @carlosdauden @CarlosRoca13 @juancarlosonate-tecnativa


carlosdauden (approved):

juancarlosonate-tecnativa (approved):
LGTM

juancarlosonate-tecnativa:
Maybe it would be worth adding a test though, and forward/backporting it.

pedrobaeza (approved):
/ocabot merge patch

